PR #51, opened 2022-03-17
WIP: Move useful functionality from monitor into a CLI package

yourbuddyconner:


anna-carroll:
nuking this PR per discussion with @yourbuddyconner 

we need to RFC / design the CLI to ensure it fits our needs 
we also need to move towards smaller, digestable/reviewable PRs instead of long-running monoPRs

also, the functionality in this PR is porting the deprecated monitoring service so we don't need it ✅

PR #54, opened 2022-03-17
fix: conflicting type packages

ErinHales:
I may be the only one experiencing this issue, but `@types/chai` and `@types/mocha` don't play nicely together and I'm unable to build or run tests if both of these packages are installed _anywhere_ in the monorepo


Imti (approved):
🦖

PR #71, opened 2022-03-23
indexer: added graphql API, cleaned up package, fixed build

yourbuddyconner:


kekonen:
I pushed new nomad versions, a small fix for getting gov domain, lots of stuff prettied

Imti (approved):
🥳 Let's merge since it's already running on prod and we can improve as we go 👌

PR #78, opened 2022-03-24
fix: multi-provider/utils.equalIds

kekonen:
Very simple fix for `equalIds` utility function.
Right now we compare 2 arrays, with `===` which is not correct and element-wise comparison is required when dealing with arrays

### DIFF
--- a/packages/multi-provider/src/utils.ts
+++ b/packages/multi-provider/src/utils.ts
@@ -100,7 +100,9 @@ export function evmId(data: BytesLike): Address {
  * @throws if either side is not a valid 20 or 32 byte nomad identifier
  */
 export function equalIds(left: BytesLike, right: BytesLike): boolean {
-  return canonizeId(left) === canonizeId(right);
+  const l = canonizeId(left);
+  const r = canonizeId(right);
+  return l.every((v, i) => v === r[i]);
 }
 
 /**


packages/multi-provider/src/utils.ts
```diff
@@ -100,7 +100,9 @@ export function evmId(data: BytesLike): Address {
  * @throws if either side is not a valid 20 or 32 byte nomad identifier
  */
 export function equalIds(left: BytesLike, right: BytesLike): boolean {
-  return canonizeId(left) === canonizeId(right);
+  const l = canonizeId(left);
+  const r = canonizeId(right);
+  return l.every((v, i) => v === r[i]);
```
prestwich, line 105: what does this fix? o.O

kekonen: @prestwich `canonizeId(...)` returns an arrayish. In JS u cannot compare 2 arrays with simple equals. Here is some more https://stackoverflow.com/questions/13523611/how-to-compare-two-arrays-in-node-js

prestwich: oh I forgot it returns a uint8array ya this is fint

kekonen: So, shall we merge?

prestwich (approved):

PR #81, opened 2022-03-24
New deploy fix + checks

kekonen:
Added checks and small fix in order to make new-deploy run


packages/new-deploy/src/DeployContext.ts
```diff
@@ -137,6 +149,7 @@ export default class DeployContext extends MultiProvider<config.Domain> {
 
     await core.deployHome();
     await core.deployGovernanceRouter();
+    // await Promise.all([core.deployHome(), core.deployGovernanceRouter()]);
```
prestwich, line 152: delete pls :)

packages/new-deploy/src/DeployContext.ts
```diff
@@ -280,6 +291,9 @@ export default class DeployContext extends MultiProvider<config.Domain> {
 
     await this.relinquish();
 
+    await this.checkCores();
```
prestwich, line 294: improvement idea for another PR:
I feel like there should be checks before `relinquish` so that we error into a recoverable state

packages/new-deploy/src/core/CoreContracts.ts
```diff
@@ -94,7 +97,7 @@ export default class EvmCoreDeploy extends AbstractCoreDeploy<config.EvmCoreCont
   }
 
   get replicas(): ReadonlyArray<string> {
-    return Object.keys(this.replicas);
+    return Object.keys(this.data.replicas!);
```
prestwich, line 100: avoid using `!` operator. instead use a `if (!whatever) throw new Error("...")`

prestwich: previously we've been using these, but they're bad hygiene and undermine the type system, so we should probably stop 😅

prestwich (changes requested):

packages/new-deploy/src/core/CoreContracts.ts
kekonen: makes total sense. fixed in [92dde9a](https://github.com/nomad-xyz/monorepo/pull/81/commits/92dde9a14ab7121d4d2abadcc316003c217d4c57)

packages/new-deploy/src/DeployContext.ts
kekonen: fixed in [92dde9a](https://github.com/nomad-xyz/monorepo/pull/81/commits/92dde9a14ab7121d4d2abadcc316003c217d4c57)

prestwich (approved):
I'm going to merge this into my branch, then change the following things:

0. lint and prettier (you should set these up to run automatically locally 😅 ) 
1. improve watcher registration flow to make more readable
2. get `replicas` should return `[]` instead of erroring

PR #83, opened 2022-03-25
Yarn version: README instructions + version upgrade

anna-carroll:
- first commit makes README to install more informative
- second commit upgrades yarn from 3.1.1 to 3.2.0


anna-carroll:
not sure if we want the second commit; can drop if desired

prestwich (approved):

PR #91, opened 2022-03-25
Deploy script bugfixes / refactoring & invocation script

anna-carroll:
- write an invocation script to run the deploy process
- patch any bugs in the deploy process found while running it
- refactor the deploy process a bit for some naming / readability


packages/new-deploy/scripts/deploy.ts
```diff
@@ -0,0 +1,34 @@
...
+
+run();
+
+/*
+* get published config
+* modify to add a network
+* run the deploy script
+* re-publish config
+* */
+const config = {}; // TODO: get published (or local??) config
```
prestwich, line 15: idea state:
- script accepts a string as input from `process.argv` or env var
- string is tried as a filepath, then a URL

packages/new-deploy/scripts/deploy.ts
```diff
@@ -0,0 +1,34 @@
...
+const DEPLOYER_PRIVATE_KEY = ""; // TODO: get from .env - also, do we need different private keys for each chain??
+const OVERRIDES = {}; // TODO write local overrides config
+
+async function run() {
+    const deployContext = new DeployContext(config);
+    for (const network of deployContext.networks) {
+        const rpc = deployContext.data.rpcs[network][0]; // TODO: note: we can't publish private RPCs
+        const provider = new ethers.providers.JsonRpcProvider(rpc);
```
prestwich, line 23: there's a shortcut for this:
`context.registerRpcProvider(rpc)`

prestwich: guess it doesn't make as much sense if you have to manually instantiate the wallet anyway 🤔

packages/new-deploy/scripts/deploy.ts
```diff
@@ -0,0 +1,34 @@
...
+        const wallet = new ethers.Wallet(DEPLOYER_PRIVATE_KEY, provider);
+        const signer = new NonceManager(wallet);
+        deployContext.registerSigner(network, signer);
+        deployContext.overrides.set(network, OVERRIDES[network]);
+    }
+    // run the deploy script
+    await deployContext.deployAndRelinquish();
+    // output the updated config
+    fs.writeFileSync(`./scripts/config.json`, JSON.stringify(deployContext.data, null, 4));
```
prestwich, line 32: TODO(James): add a function for this to the deploy itself

packages/new-deploy/scripts/deploy.ts
anna-carroll: hmm but you're right it should actually be calling `context.registerRpcProvider(rpc)` when it constructs the `DeployContext`. i'll do `deployContext.mustGetProvider` instead of instantiating it separately.

anna-carroll: https://github.com/nomad-xyz/monorepo/pull/91/commits/69583fc3a930a74fbfe5d758a9c58f6d8d1f92be

packages/new-deploy/src/DeployContext.ts
```diff
@@ -113,29 +113,97 @@ export default class DeployContext extends MultiProvider<config.Domain> {
...
+
+    // appoint governor on all networks
+    await Promise.all(
+      this.networks.map((network) =>
+        this.mustGetCore(network).appointGovernor(),
+      ),
+    );
+
+    // perform validation checks on core and bridges
+    await this.checkCores();
```
anna-carroll, line 153: TODO: this should not be called in the same process as the deploy script. in production, the governance txs need to go through before the checks can pass. and this fn should return the governance txs, rather than throwing

packages/new-deploy/src/bridge/BridgeContracts.ts
```diff
@@ -456,25 +491,6 @@ export default class BridgeContracts extends AbstractBridgeDeploy<config.EvmBrid
 
     // check verification addresses
     // TODO: add beacon and proxy where needed.
-    this.checkVerificationInput(
```
anna-carroll, line 459: the full verification input doesn't exist every time the deploy script runs. it only exists fully when you deploy everything fresh.

packages/new-deploy/src/bridge/BridgeContracts.ts
```diff
@@ -314,12 +329,12 @@ export default class BridgeContracts extends AbstractBridgeDeploy<config.EvmBrid
     return [];
   }
 
-  async deployCustomTokens(): Promise<void> {
+  async deployCustomTokens(): Promise<ethers.PopulatedTransaction[][]> {
```
anna-carroll, line 332: this is not actually called anywhere.

ideally, it would be called before relinquishing the bridge, so that in the case we're deploying a fresh network, this happens at deploy-time before relinquishing the bridge (and before any tokens are sent over the bridge)

packages/new-deploy/src/DeployContext.ts
```diff
@@ -113,29 +113,97 @@ export default class DeployContext extends MultiProvider<config.Domain> {
...
+      networksToDeploy.map((net) =>
+        this.deployCore(this.mustGetDomainConfig(net)),
+      ),
+    );
   }
 
   protected async deployCore(domain: config.Domain): Promise<void> {
+    console.log('deployCore ', domain);
+
     this.addDomain(domain);
```
anna-carroll, line 207: is `addDomain` calls `config.addNetwork` - (1) is this fn idempotent? (2) shouldn't the domain already be added if this fn is being called?

packages/new-deploy/src/DeployContext.ts
```diff
@@ -155,32 +221,29 @@ export default class DeployContext extends MultiProvider<config.Domain> {
     this.addCore(domain.name, complete);
   }
 
-  // post-deployment setup
-  // We choose to enroll all currently known routers and watchers here,
-  // event those that may not be immediately connected, so that IF they are
-  // connected in the future, the routers will be properly configured.
-  // Essentially we run likely future governance actions at deploy time, so
-  // we will not be required to run them later.
-  protected async enrollCores(): Promise<ethers.PopulatedTransaction[]> {
```
anna-carroll, line 164: why are some fns in the deploy process protected and some lower down in the call stack aren't?

packages/new-deploy/src/DeployContext.ts
```diff
@@ -113,26 +113,92 @@ export default class DeployContext extends MultiProvider<config.Domain> {
     this._data = config.addBridge(this.data, name, bridge);
   }
 
-  pushVerification(
```
anna-carroll, line 116: this is added back at the bottom, just moving things around in the file don't mind me 😬

packages/new-deploy/src/bridge/BridgeContracts.ts
anna-carroll: as a higher-level comment, we need to ensure we are outputting the verification input, ideally in the same format which can be consumed by our existing verification code.

anna-carroll: https://github.com/nomad-xyz/monorepo/issues/103

packages/new-deploy/src/bridge/BridgeContracts.ts
anna-carroll: https://github.com/nomad-xyz/monorepo/issues/104

prestwich: flatten the `[][]` to `ethers.PopulatedTransaction[]`

prestwich: with `.flat()`

anna-carroll: this is not perfect cuz everywhere else deploys in one fn then enrolls in another; whereas deployCustomToken both deploys and enrolls.
but rocking with it for now: https://github.com/nomad-xyz/monorepo/pull/91/commits/4139b225aeb0bbc18ff70e49fbc10eab5ccc9cd0

packages/new-deploy/scripts/deploy.ts
anna-carroll: - implemented local filepath
- added ticket for attempting URL: https://github.com/nomad-xyz/monorepo/issues/112

packages/new-deploy/package.json
```diff
@@ -24,12 +24,14 @@
     "build": "tsc --build",
     "check": "tsc --noEmit",
     "prettier": "prettier --write ./src",
-    "lint": "eslint --fix ./src"
+    "lint": "eslint --fix ./src",
+    "deploy": "ts-node scripts/deploy.ts config/development/config.json config/development/overrides.json",
+    "check-deploy": "ts-node scripts/check.ts config/development/config.json"
```
prestwich, line 29: document how to invoke these and what their arguments are instead of making yarn script aliases

packages/new-deploy/src/DeployContext.ts
prestwich: reordering makes reviewing changes near-impossible. move it back and reorder in a separate PR

prestwich (changes requested):

packages/new-deploy/package.json
anna-carroll: https://github.com/nomad-xyz/monorepo/issues/140

PR #139, opened 2022-04-05
refactor: callbatch instead of tx array

### LINKED ISSUE #111: fix: output governance actions in usable way
governance actions are pushed to a generic array, not associated with which chain they need to be sent to - so they can't really be executed at the moment

I think we should implement the "ideal state" documented here: https://github.com/nomad-xyz/monorepo/discussions/109#discussioncomment-2485792

use `CallBatch` stored on `DeployContext`

anna-carroll:
closes #111

### DIFF
--- a/packages/multi-provider/src/index.ts
+++ b/packages/multi-provider/src/index.ts
@@ -3,4 +3,11 @@ export { Contracts } from './contracts';
 
 export * as utils from './utils';
 
+export {
+  UnreachableError,
+  WithContext,
+  NoProviderError,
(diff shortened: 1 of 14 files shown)


anna-carroll (commented):
this PR is very beautiful to me 🥲 

I ran it with the deploy script and confirmed it's outputting all the right governance txs!!!! HUGE

the only problem is that it throws when trying to build the governance batch. i submitted a ticket with details here: #146 

going to merge as-is because it works!!

PR #154, opened 2022-04-07
feature: invocation script for governance transaction execution

anna-carroll:
## Motivation

Need a way to execute governance transactions on governor chain and remote chains

## Solution

Write a script utilizing `sdk-govern` to invoke built-in `execute` and `executeDomain` functions

## PR Checklist

- [x] Added Tests
- [x] Updated Documentation

**TODOs:**
- [x] implement overrides. - `MultiProvider` `setOverrides`?
- [x] remove package.json scripts, replace with documentation


anna-carroll:
NOTE: convenience script to pipe deploy outputs (config and callbatch) into governance invocation?

packages/new-deploy/src/DeployContext.ts
```diff
@@ -35,12 +35,7 @@ export class DeployContext extends MultiProvider<config.Domain> {
   }
 
   get asNomadContext(): NomadContext {
-    const ctx = new NomadContext(this.data);
-
-    for (const [domain, provider] of this.providers.entries()) {
-      ctx.registerProvider(domain, provider);
-    }
-    return ctx;
+    return new NomadContext(this.data);
```
prestwich, line 38: this will break things no?

anna-carroll: i don't think so - registering provider is moved into the `NomadContext` constructor, so it's called there

packages/sdk-govern/src/index.ts
```diff
@@ -27,6 +27,25 @@ export interface CallBatchContents {
   built?: ethers.PopulatedTransaction;
 }
 
+export enum BatchStatus {
+  NOT_RECEIVED,
+  RECEIVED,
+  EXECUTED,
+}
+
+export function BatchStatusText(status: number): string {
```
anna-carroll, line 36: this might be whack naming, maybe should move to camelCase

packages/new-deploy/src/DeployContext.ts
anna-carroll: i suppose it's feasible that the DeployContext has other / more providers registered that aren't in the `data`?

prestwich: yes exactly. plus, signers aren't registered in the constructor

packages/sdk-govern/src/index.ts
prestwich: use a string enum here instead of the function

`https://www.typescriptlang.org/docs/handbook/enums.html#string-enums`

anna-carroll:
**think about:**
- [ ] reduce boiler plate for invocation scripts
   - [ ] load config/overrides
   - [ ] instantiate context
   - [ ] register signers / providers
   - [ ] set overrides
- [ ] convenience script to pipe deploy outputs from `new-deploy` (config and callbatch) into governance invocation?

packages/new-deploy/src/DeployContext.ts
anna-carroll: https://github.com/nomad-xyz/monorepo/pull/154/commits/058b8788f36e8655af5fc7f559d3445f711ccdd5

packages/sdk-govern/src/index.ts
prestwich: Edit: I mean a reverse mapping

https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings

anna-carroll: https://github.com/nomad-xyz/monorepo/pull/154/commits/8048bd78140c62c0ab53905ebe7e5258f12210d5

packages/sdk/tests/index.test.ts
```diff
@@ -64,14 +63,6 @@ describe('sdk', async () => {
       }
     });
 
-    it('errors if signer is registered without a provider', () => {
```
prestwich, line 67: where did these go?

packages/sdk-govern/scripts/executeGovernorDomain.ts
```diff
@@ -0,0 +1,52 @@
+import * as config from '@nomad-xyz/configuration';
+import {NomadContext} from "@nomad-xyz/sdk";
+import {getCallBatch, getConfig} from './utils';
+import {CallBatch, CallBatchContents} from "../dist";
+import * as dotenv from 'dotenv';
+import {equalIds} from "@nomad-xyz/multi-provider/dist/utils";
+import * as ethers from "ethers";
+dotenv.config();
+
+run();
```
prestwich, line 10: if we're adding 3 extra CLIs it's time to just suck it up and write a [commander](https://www.npmjs.com/package/commander) cli

packages/sdk-bridge/src/BridgeContext.ts
```diff
@@ -22,16 +22,11 @@ export class BridgeContext extends NomadContext {
...
     super(environment);
-
     this.bridges = new Map();
-    const bridges = this.conf.networks.map(
-      (network) =>
-        new BridgeContracts(this, network, this.conf.bridge[network]),
-    );
-
-    bridges.forEach((bridge) => {
+    for (const network of this.conf.networks) {
```
prestwich, line 26: why driveby changes?

packages/new-deploy/src/DeployContext.ts
```diff
@@ -42,7 +42,6 @@ export class DeployContext extends MultiProvider<config.Domain> {
 
   get asNomadContext(): NomadContext {
     const ctx = new NomadContext(this.data);
-
```
prestwich, line 45: why driveby changes?

prestwich (approved):
approving. some cleanup could be done as separate PRs. the script approach is getting unwieldy and the usage of argv is brittle

packages/sdk/tests/index.test.ts
anna-carroll: it no longer errors because it has a default provider instantiated

packages/sdk-govern/scripts/executeGovernorDomain.ts
anna-carroll: https://github.com/nomad-xyz/monorepo/issues/243

PR #155, opened 2022-04-08
bug: get replica name or domain

### LINKED ISSUE #156: bug: getReplica only accepts domain
`getReplica` should accept `nameOrDomain` and `resolveDomainName`

ErinHales:
Closes #156

`getReplica` on NomadContext only accepts domain number where everything else accepts nameOrDomain

## Solution

Pass in nameOrDomain, resolveDomain

## PR Checklist

- [x] Added Tests
- [x] Updated Documentation

### DIFF
--- a/packages/sdk/src/CoreContracts.ts
+++ b/packages/sdk/src/CoreContracts.ts
@@ -29,6 +29,12 @@ export class CoreContracts<T extends NomadContext> extends Contracts<
     this.conf = conf;
   }
 
+  /**
+   * Resolves Replica with given domain
+   *
+   * @param nameOrDomain The name or domain ID of the Replica
+   * @returns An interface for the Replica, undefined if not found
+   */
   getReplica(nameOrDomain: string | number): core.Replica | undefined {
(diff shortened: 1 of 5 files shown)


packages/sdk/src/CoreContracts.ts
```diff
@@ -29,7 +29,8 @@ export class CoreContracts<T extends NomadContext> extends Contracts<
     this.conf = conf;
   }
 
-  getReplica(domain: string): core.Replica | undefined {
+  getReplica(nameOrDomain: string | number): core.Replica | undefined {
+    const domain = this.context.resolveDomain(nameOrDomain);
     if (!this.connection) throw new NoProviderError(this.context, domain);
     const replica = this.conf.replicas[domain];
```
prestwich, line 41: why is this not a type error?

`domain` is a `number`
`this.conf.replicas` is a `Record<string, _>`

why can we index it with a number? this is a bug. Need to use `resolveDomainName` on L33

prestwich (changes requested):

prestwich (approved):

Issue #103, opened 2022-03-31
fix: output contract verification information in compatible format


anna-carroll:
"compatible format" depends on the code written in #124 

- in the past it was an array of verification inputs, one file per chain
- presently, the deploy script creates one file for all chains - a map of network -> array of verification inputs

yourbuddyconner:
Request: 

Based on the discussion in #124, we're solidifying on the pattern of verifying the `core` and `bridge` contracts independently of one-another. The ideal format for the `verification.json` file is as follows: 

``` json
{
    "<network>": {
        "core": {
            // CORE CONTRACT VERIFICATION INPUTS
        }, 
        "bridge": {
            // BRIDGE CONTRACT VERIFICATION INPUTS
        }
    }
}
```

PR #157, opened 2022-04-08
Foundry-based etherscan verification

### LINKED ISSUE #124: feat: make package for verifying contracts (port from old deploy)

prestwich:
closes #124 and #103

Foundry-based contract verification automation

## Motivation

Contract verification using hardhat etherscan has caused some brittle code to be written. We'd like to improve it :)

## Solution

- disable metadata in hh solc
- Ensure that hh and forge are configured to produce byte-for-byte identical output
- improve our JSON verification output format
- build a workflow based on `forge verify-contract` using our JSON verification output
- provide a script that can be run from `packages/new-deploy` to verify all contracts at once

I considered making a plugin that extends hh's etherscan-verify, and decided it was easier to extend foundry.

## PR Checklist

- [ ] Added Tests
- [x] Updated Documentation


packages/new-deploy/src/bridge/BridgeContracts.ts
```diff
@@ -170,14 +184,25 @@ export default class BridgeContracts extends AbstractBridgeDeploy<config.EvmBrid
     const ubc = core.upgradeBeaconController.address;
     if (!ubc) throw new Error('Cannot deploy proxy without UBC');
     const factory = new UpgradeBeacon__factory(this.context.getDeployer(name));
-    const beacon = await factory.deploy(implementation, ubc, this.overrides);
+
+    const constructorArguments = [implementation, ubc];
+    const beacon = await factory.deploy(
+      constructorArguments[0],
```
anna-carroll, line 190: could do `...constructorArguments` so the code is the same no matter how many args there are

packages/new-deploy/src/core/CoreContracts.ts
```diff
@@ -253,14 +273,25 @@ export default class EvmCoreDeploy extends AbstractCoreDeploy<config.EvmCoreCont
     const factory = new UpgradeBeaconProxy__factory(
       this.context.getDeployer(name),
     );
-    const prx = await factory.deploy(beacon, initData, this.overrides);
+    const constructorArguments = [beacon, initData];
+    const prx = await factory.deploy(
+      beacon, // constructorArguments[0],
```
anna-carroll, line 278: fix?

packages/new-deploy/scripts/verify.ts
```diff
@@ -0,0 +1,245 @@
+// --chain-id
```
anna-carroll, line 1: can you comment some context for what this is

packages/new-deploy/scripts/verify.ts
```diff
@@ -0,0 +1,245 @@
...
+const execAsync = promisify(exec);
+const readFileAsync = promisify(readFile);
+const writeFileAsync = promisify(writeFile);
+
+/// async sleep function
+function delay(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+const etherscanKey = process.env.ETHERSCAN_KEY;
```
anna-carroll, line 32: can you add an error which throws informatively if there's no API key present

packages/new-deploy/scripts/verify.ts
```diff
@@ -0,0 +1,245 @@
...
+
+/// Extract the etherscan GUID from the `forge verify-contract` stdout
+export function extractGuidFromFoundry(res: {
+  stdout: string;
+  stderr: string;
+}): string | undefined {
+  if (res.stderr.indexOf('already verified') !== -1) return;
+  const bits = res.stdout.split('`');
+  if (bits.length < 4)
```
anna-carroll, line 41: ideally could just add some comments 
```
// we expect there to be 4 arguments
// the GUID is the 4th argument
```

packages/new-deploy/scripts/verify.ts
```diff
@@ -0,0 +1,245 @@
...
+      this.verified = true;
+    }
+  }
+
+  toJSON(): VerificationOutput {
+    return this._data;
+  }
+}
+
+async function run() {
```
anna-carroll, line 198: hmmm in my mind, ideally we could split this up into a diff file than the class?

packages/new-deploy/scripts/verify.ts
```diff
@@ -0,0 +1,245 @@
...
+  console.log();
```
anna-carroll, line 199: remove

packages/new-deploy/scripts/verify.ts
```diff
@@ -0,0 +1,245 @@
...
+  const environment = process.argv[2];
```
anna-carroll, line 200: imo we should remove this as an arg and use the `config` stored locally, no?

anna-carroll (approved):
generally, LGTM. left some comments. going to merge so i can use the new format for the new deploy.

packages/new-deploy/src/bridge/BridgeContracts.ts
```diff
@@ -146,14 +149,25 @@ export default class BridgeContracts extends AbstractBridgeDeploy<config.EvmBrid
     const factory = new UpgradeBeaconProxy__factory(
       this.context.getDeployer(name),
     );
-    const prx = await factory.deploy(beacon, initData, this.overrides);
+
+    const constructorArguments = [beacon, initData];
+    const prx = await factory.deploy(
+      beacon, //constructorArguments[0],
```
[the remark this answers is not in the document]
prestwich: that's me being lazy with types. `constructorArguments[]` contains `ethers.BytesLike` which is less specific than `beacon: string`. so casting from an element of the args array to string is not safe from the typesystem's perspective. rather than clutter up the logic with casts, I decided to leave a not for future james

packages/new-deploy/src/bridge/BridgeContracts.ts
prestwich: doesn't work. spread operator only works for rest args :(

packages/new-deploy/scripts/verify.ts
prestwich: next PR.

packages/new-deploy/scripts/verify.ts
prestwich: I'll leave a comment saying that it's lazy and brittle lol. next PR.

packages/new-deploy/scripts/verify.ts
prestwich: the future change should be to incorporate the verification class into the deploy script. instead of doing the intermediate change, I'd like to just leave it how it is until we do the proper fix. a mono-file script is fine if messy.

packages/new-deploy/scripts/verify.ts
prestwich: intentional line break

packages/new-deploy/scripts/verify.ts
prestwich: next PR

PR #195, opened 2022-04-13
feature: restart-safe deploys

### LINKED ISSUE #115: feat: restart-safe deploys
Right now, if the deploy script throws at any point in the process, you have to restart from zero

Thus far, the deploy script has been prone to some instability due to 
- RPC endpoint rate limiting the code (hard to fix)
- too slow gas price or too low gas limit set in overrides (fixable by hard-coding high values, though this kinda sucks in production)
And a myriad of other potential developer errors or ephemeral problems like internet or network instability

It would be dope if you could re-run the deploy script when it fails and it would pick up where it left off

anna-carroll:
## Motivation

when deploy fails, you have to start from scratch again. we want to save the deployed contracts in the config and the verification inputs so we can restart gracefully

closes #115 

## Solution

wrap the deploy invocation in a try/catch and cache the config and verification when it fails

## PR Checklist

- [x] FIX: re-running the script clobbers old verification outputs
- [x] FIX: writing empty verification inputs
- [x] FIX: writing empty governance transactions
- [ ] FIX: partial configs throw with rust verification #216 
- [ ] Added Tests
- [x] Updated Documentation

### DIFF
--- a/packages/new-deploy/README.md
+++ b/packages/new-deploy/README.md
@@ -9,8 +9,8 @@ process based on the Nomad configuration format.
 This package is feature-complete and we have big plans for the future :)
 
 - Run a fresh deploy with no networks
-- Extend a current deployment
-- Future: gracefully resume partial deployments
+- Extend a current deployment with new networks or connections
+- Gracefully resume partial deployments by re-running the script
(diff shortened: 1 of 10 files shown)


kekonen (approved):
lgtm

PR #211, opened 2022-04-18
fix: token registry owner

anna-carroll:
## Motivation

Deploy script transfers ownership of token registry to governance; it should be transferred to bridge router

## Solution

- modify deploy script to transfer ownership fo token registry to bridge router
- modify deploy script checks to check the correct state 

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/new-deploy/src/bridge/BridgeContracts.ts
+++ b/packages/new-deploy/src/bridge/BridgeContracts.ts
@@ -602,8 +602,9 @@ export default class BridgeContracts extends AbstractBridgeDeploy<config.EvmBrid
     assertBeaconProxy(this.data.tokenRegistry, 'TokenRegistry');
     // owner
     const tokenRegistryOwner = await this.tokenRegistryContract.owner();
-    expect(utils.equalIds(tokenRegistryOwner, core.governanceRouter.address)).to
-      .be.true;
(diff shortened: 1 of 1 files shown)


packages/new-deploy/src/bridge/BridgeContracts.ts
```diff
@@ -524,25 +524,23 @@ export default class BridgeContracts extends AbstractBridgeDeploy<config.EvmBrid
...
-        const owner = await contract.owner();
-        if (utils.equalIds(owner, deployer)) {
-          log(`transfer bridge ownership on ${name}`);
-          return await contract.transferOwnership(governance, this.overrides);
-        }
-      }),
-    );
+    // transfer ownership of tokenRegistry to bridge
+    const registryOwner = await this.tokenRegistryContract.owner();
+    if (utils.equalIds(registryOwner, deployer)) {
```
prestwich, line 531: is this pattern used everywhere? it fails if `deployer` is equal to (e.g.) the governance because we have used a safe-ethers-signer. File an issue to correct this.

correct pattern would be `if (!utils.equalIds(registryOwner, correctOwner)))`

prestwich (approved):
changes LGTM. please open the new issue noted

PR #222, opened 2022-04-20
add: process method on NomadContext

### LINKED ISSUE #134: sdk-bridge: add convenience method BridgeContext.proveAndProcess()
Currently, in the GUI, we have to extract the appropriate `replica` contract, and call `replica.proveAndProcess()` on it [like so](https://github.com/nomad-xyz/nomad-app/blob/963e89e675f666a145966b1b81bd5f6f78c78e73/src/store/modules/sdk.ts#L251-L280). 

We should add a `NomadContext.proveAndProcess(message: NomadMessage)` method which allows for easy processing of a message. 

The method should: 
- Fetch the required proof from S3
- Unpack it and call `replica.proveAndProcess` on the correct replica based on `message.origin`
- Throw a helpful error if it's not time to process yet

ErinHales:
Closes #134 

Adds process method to the sdk

## Motivation

We should make this more accessible for other applications using our sdk. Decided to add to NomadContext so any messages could be processed, not just Bridge messages.

## Solution

 - Fetch transaction details from Indexer API
 - Fetch proof from S3
 - Ensure signer is registered
 - Call `proveAndProcess` on Replica

## PR Checklist

- [ ] Added Tests
- [x] Updated Documentation

### DIFF
--- a/packages/sdk/src/NomadContext.ts
+++ b/packages/sdk/src/NomadContext.ts
@@ -1,4 +1,4 @@
-import { providers, Signer } from 'ethers';
+import { providers, Signer, BigNumber, ContractTransaction } from 'ethers';
 
 import { MultiProvider } from '@nomad-xyz/multi-provider';
 import * as core from '@nomad-xyz/contracts-core';
@@ -172,6 +172,43 @@ export class NomadContext extends MultiProvider<config.Domain> {
    */
   governorCore(): CoreContracts<this> {
     return this.mustGetCore(this.governor.domain);
+  }
+
+  async process(txId: string): Promise<ContractTransaction>{
+    const isProduction = this.conf.environment === 'production'
+    const nomadAPI = isProduction
+      ? 'https://bridge-indexer.prod.madlads.tools/tx/'
+      : 'https://bridge-indexer.dev.madlads.tools/tx/'
+    const s3URL = isProduction
+      ? 'https://nomadxyz-production-proofs.s3.us-west-2.amazonaws.com/'
+      : 'https://nomadxyz-development-proofs.s3.us-west-2.amazonaws.com/'
+
+    // get transfer message
+    const res = await fetch(`${nomadAPI}${txId}`)
+    const tx = (await res.json())[0] as any
+
+    // get proof
+    const index = BigNumber.from(tx.leafIndex).toNumber()
+    const originName = this.resolveDomainName(tx.origin)
+    const s3Res = await fetch(`${s3URL}${originName}_${index}`)
+    const data = (await s3Res.json()) as any
+    console.log('proof: ', data)
+
+    // get replica contract
+    const replica = this.getReplicaFor(tx.origin, tx.destination)
+
+    if (!replica) throw new Error('missing replica, unable to process transaction')
+
+    // get signer and connect replica
+    const signer = this.getSigner(tx.destination)
+    if (!signer) throw new Error('missing signer, unable to process transaction')
+    replica.connect(signer)
+
+    return await replica.proveAndProcess(
+      data.message,
+      data.proof.path,
+      data.proof.index
+    )
   }
 
   blacklist(): Set<number> {
--- a/packages/sdk/src/NomadContext.ts
+++ b/packages/sdk/src/NomadContext.ts
@@ -217,10 +217,6 @@ export class NomadContext extends MultiProvider<config.Domain> {
     // get replica contract
     const replica = this.mustGetReplicaFor(originNetwork, destNetwork);
 
-    // get signer and connect replica
-    const signer = this.mustGetSigner(destNetwork);
-    replica.connect(signer);
-
     return replica.proveAndProcess(
       data.message,
       data.proof.path,
--- a/packages/sdk/src/NomadContext.ts
+++ b/packages/sdk/src/NomadContext.ts
@@ -5,7 +5,7 @@ import * as core from '@nomad-xyz/contracts-core';
 import * as config from '@nomad-xyz/configuration';
 
 import { CoreContracts } from './CoreContracts';
-import { NomadMessage } from './messages/NomadMessage'
+import { NomadMessage } from './messages/NomadMessage';
 
 export type Address = string;
 
@@ -58,7 +58,7 @@ export class NomadContext extends MultiProvider<config.Domain> {
         this.registerRpcProvider(network, this.conf.rpcs[network][0]);
       }
       // set core contracts
-      const core = new CoreContracts(this, network, this.conf.core[network])
+      const core = new CoreContracts(this, network, this.conf.core[network]);
       this._cores.set(core.domain, core);
     }
   }
@@ -68,7 +68,7 @@ export class NomadContext extends MultiProvider<config.Domain> {
   }
 
   get environment(): string {
-    return this.conf.environment
+    return this.conf.environment;
   }
 
   /**
@@ -205,27 +205,27 @@ export class NomadContext extends MultiProvider<config.Domain> {
    * @returns The Contract Transaction receipt
    */
   async process(message: NomadMessage<NomadContext>): Promise<ContractTransaction>{
-    const s3URL = `https://nomadxyz-${this.environment}-proofs.s3.us-west-2.amazonaws.com/`
+    const s3URL = `https://nomadxyz-${this.environment}-proofs.s3.us-west-2.amazonaws.com/`;
 
-    const originNetwork = this.resolveDomainName(message.origin)
-    const destNetwork = this.resolveDomainName(message.destination)
-    const index = message.leafIndex.toString()
-    const s3Res = await fetch(`${s3URL}${originNetwork}_${index}`)
-    if (!s3Res) throw new Error('Not able to fetch proof')
-    const data: MessageProof = await s3Res.json()
+    const originNetwork = this.resolveDomainName(message.origin);
+    const destNetwork = this.resolveDomainName(message.destination);
+    const index = message.leafIndex.toString();
+    const s3Res = await fetch(`${s3URL}${originNetwork}_${index}`);
+    if (!s3Res) throw new Error('Not able to fetch proof');
+    const data: MessageProof = await s3Res.json();
 
     // get replica contract
-    const replica = this.mustGetReplicaFor(originNetwork, destNetwork)
+    const replica = this.mustGetReplicaFor(originNetwork, destNetwork);
 
     // get signer and connect replica
-    const signer = this.mustGetSigner(destNetwork)
-    replica.connect(signer)
+    const signer = this.mustGetSigner(destNetwork);
+    replica.connect(signer);
 
     return replica.proveAndProcess(
       data.message,
       data.proof.path,
       data.proof.index
-    )
+    );
   }
 
   blacklist(): Set<number> {
--- a/packages/sdk/src/NomadContext.ts
+++ b/packages/sdk/src/NomadContext.ts
@@ -211,6 +211,7 @@ export class NomadContext extends MultiProvider<config.Domain> {
     const destNetwork = this.resolveDomainName(message.destination)
     const index = message.leafIndex.toString()
     const s3Res = await fetch(`${s3URL}${originNetwork}_${index}`)
+    if (!s3Res) throw new Error('Not able to fetch proof')
     const data: MessageProof = await s3Res.json()
 
     // get replica contract
--- a/packages/sdk/src/NomadContext.ts
+++ b/packages/sdk/src/NomadContext.ts
@@ -1,4 +1,4 @@
-import { providers, Signer, ContractTransaction } from 'ethers';
+import { providers, Signer, ContractTransaction, BytesLike } from 'ethers';
 
 import { MultiProvider } from '@nomad-xyz/multi-provider';
 import * as core from '@nomad-xyz/contracts-core';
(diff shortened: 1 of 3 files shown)


packages/sdk/src/NomadContext.ts
```diff
@@ -173,6 +173,43 @@ export class NomadContext extends MultiProvider<config.Domain> {
...
+    const s3URL = isProduction
+      ? 'https://nomadxyz-production-proofs.s3.us-west-2.amazonaws.com/'
+      : 'https://nomadxyz-development-proofs.s3.us-west-2.amazonaws.com/'
+
+    // get transfer message
+    const res = await fetch(`${nomadAPI}${txId}`)
+    const tx = (await res.json())[0] as any
+
+    // get proof
+    const index = BigNumber.from(tx.leafIndex).toNumber()
```
anna-carroll, line 190: i think SDK should query on-chain, not data from indexer. i don't think we have enough load-testing to make indexer stable for external queries. all of the transaction information, though, is available in the `NomadMessage` object in the SDK

packages/sdk/src/NomadContext.ts
```diff
@@ -173,6 +173,43 @@ export class NomadContext extends MultiProvider<config.Domain> {
     return this.mustGetCore(this.governor.domain);
   }
 
+  async process(txId: string): Promise<ContractTransaction>{
+    const isProduction = this.conf.environment === 'production'
+    const nomadAPI = isProduction
+      ? 'https://bridge-indexer.prod.madlads.tools/tx/'
+      : 'https://bridge-indexer.dev.madlads.tools/tx/'
+    const s3URL = isProduction
```
anna-carroll, line 181: have we discussed S3 proofs as being okay to serve externally?

anna-carroll (commented):
i think we should swap indexer for on-chain calls here :)

packages/sdk/src/NomadContext.ts
ErinHales: I'll check

prestwich: We already do for self-proving, no?

packages/sdk/src/NomadContext.ts
prestwich: SDK needs to query indexer eventually. the chain event querying is not sufficient and leads to the terrible kludges we have. Are you arguing that now is not the time?

packages/sdk/src/NomadContext.ts
```diff
@@ -173,6 +173,43 @@ export class NomadContext extends MultiProvider<config.Domain> {
...
+      ? 'https://nomadxyz-production-proofs.s3.us-west-2.amazonaws.com/'
+      : 'https://nomadxyz-development-proofs.s3.us-west-2.amazonaws.com/'
+
+    // get transfer message
+    const res = await fetch(`${nomadAPI}${txId}`)
+    const tx = (await res.json())[0] as any
```
prestwich, line 187: no use of `any`

packages/sdk/src/NomadContext.ts
```diff
@@ -1,13 +1,20 @@
...
 import { MultiProvider } from '@nomad-xyz/multi-provider';
 import * as core from '@nomad-xyz/contracts-core';
 import * as config from '@nomad-xyz/configuration';
 
 import { CoreContracts } from './CoreContracts';
+import { NomadMessage } from './messages/NomadMessage'
 
 export type Address = string;
 
+const s3URLs: { [key: string]: string } = {
```
anna-carroll, line 12: could just do 
`https://nomadxyz-${environment}-proofs.s3.us-west-2.amazonaws.com/`
but would be more brittle if the URLs did change in the future

packages/sdk/src/NomadContext.ts
```diff
@@ -173,6 +180,51 @@ export class NomadContext extends MultiProvider<config.Domain> {
...
+   * 
+   * @dev Ensure that a transaction is ready to be processed. You should ensure the following
+   * criteria have been met prior to calling this function:
+   *  1. the tx has been relayed
+   *  2. the `confirmAt` timestamp for the tx is in the past
+   *  3. the user is on the destination chain in their wallet
+   *
+   * @returns The Contract Transaction receipt
+   */
+  async process(origin: string | number, txId: string): Promise<ContractTransaction>{
```
anna-carroll, line 195: is `txId` = `transactionHash`?
can we document function parameters, and consider renaming param to `transactionHash` for clarity (not sure if it's a Nomad-specific ID for transactions, for example)

packages/sdk/src/NomadContext.ts
```diff
@@ -173,6 +180,51 @@ export class NomadContext extends MultiProvider<config.Domain> {
...
 
+  /**
+   * Proves and Processes a transaction on the destination chain. This is subsidize and
+   * automatic on non-Ethereum destinations
+   * 
+   * @dev Ensure that a transaction is ready to be processed. You should ensure the following
+   * criteria have been met prior to calling this function:
+   *  1. the tx has been relayed
+   *  2. the `confirmAt` timestamp for the tx is in the past
+   *  3. the user is on the destination chain in their wallet
```
anna-carroll, line 191: this is a more GUI-facing thing I believe, within the context of the SDK, providers and signers are set on the Context object, so #3 is not relevant

packages/sdk/src/NomadContext.ts
```diff
@@ -173,6 +180,51 @@ export class NomadContext extends MultiProvider<config.Domain> {
...
   }
 
+  /**
+   * Proves and Processes a transaction on the destination chain. This is subsidize and
+   * automatic on non-Ethereum destinations
+   * 
+   * @dev Ensure that a transaction is ready to be processed. You should ensure the following
+   * criteria have been met prior to calling this function:
+   *  1. the tx has been relayed
+   *  2. the `confirmAt` timestamp for the tx is in the past
```
anna-carroll, line 190: for 1 and 2 we should include context on how to check these things using the SDK

packages/sdk/src/NomadContext.ts
```diff
@@ -173,6 +180,51 @@ export class NomadContext extends MultiProvider<config.Domain> {
...
+   *  3. the user is on the destination chain in their wallet
+   *
+   * @returns The Contract Transaction receipt
+   */
+  async process(origin: string | number, txId: string): Promise<ContractTransaction>{
+    // get s3 proof URL for environment
+    const { environment } = this.conf
```
anna-carroll, line 197: add a getter `this.environment`

packages/sdk/src/NomadContext.ts
```diff
@@ -173,6 +180,51 @@ export class NomadContext extends MultiProvider<config.Domain> {
...
+    const s3URL = s3URLs[environment]
+
+    const message = (await NomadMessage.baseFromTransactionHash(
```
anna-carroll, line 200: hmmmm maybe we add some version of `singleFromTransactionHash` that doesn't throw if there are multiple messages? `firstFromTransactionHash`? hmmmmmmmmmmmmmmmmmmm

actually maybe even better, this function takes in a `NomadMessage` instead of an `origin` and `txId`?

packages/sdk/src/NomadContext.ts
```diff
@@ -173,6 +180,51 @@ export class NomadContext extends MultiProvider<config.Domain> {
...
+      this,
+      origin,
+      txId
+    ))[0]
+
+    const originNetwork = this.resolveDomainName(message.origin)
+    const destNetwork = this.resolveDomainName(message.destination)
+    const index = message.leafIndex.toString()
+    const s3Res = await fetch(`${s3URL}${originNetwork}_${index}`)
```
anna-carroll, line 209: could extract a method `getProof` which  does

    const s3URL = s3URLs[this.environment]
    const s3Res = await fetch(`${s3URL}${originNetwork}_${index}`)

packages/sdk/src/NomadContext.ts
```diff
@@ -173,6 +180,51 @@ export class NomadContext extends MultiProvider<config.Domain> {
...
+    const data = (await s3Res.json()) as any
+
+    // get replica contract
+    const replica = this.getReplicaFor(originNetwork, destNetwork)
+    if (!replica) throw new Error('missing replica, unable to process transaction')
+
+    // get signer and connect replica
+    const signer = this.getSigner(destNetwork)
+    if (!signer) throw new Error('missing signer, unable to process transaction')
+    replica.connect(signer)
```
anna-carroll, line 219: does `getReplicaFor` not connect the signer already? (note to self: check)

anna-carroll: there should be `mustGetReplicaFor` and `mustGetSigner`, i think?

packages/sdk/src/NomadContext.ts
```diff
@@ -173,6 +180,51 @@ export class NomadContext extends MultiProvider<config.Domain> {
...
+
+    return await replica.proveAndProcess(
```
anna-carroll, line 221: nit: don't need to `await`

packages/sdk/src/NomadContext.ts
Imti: Yep, it's already being served externally from the gui (https://github.com/nomad-xyz/nomad-app/blob/development/src/config/index.ts#L19)

packages/sdk/src/NomadContext.ts
```diff
@@ -173,6 +180,51 @@ export class NomadContext extends MultiProvider<config.Domain> {
...
+      this,
+      origin,
+      txId
+    ))[0]
+
+    const originNetwork = this.resolveDomainName(message.origin)
+    const destNetwork = this.resolveDomainName(message.destination)
+    const index = message.leafIndex.toString()
+    const s3Res = await fetch(`${s3URL}${originNetwork}_${index}`)
+    const data = (await s3Res.json()) as any
```
Imti, line 210: You can probably remove any just fine here (and the surrounding curly braces)

packages/sdk/src/NomadContext.ts
anna-carroll: probably not the time, given 
- the indexer is running from a feature branch; it hasn't been finished or merged to main 
- it's been live running in production for a week 
- there is no documentation 
- there is no monitoring / tracing to debug issues in the code
- there is one engineer who has a deep understanding of the code

packages/sdk/src/NomadContext.ts
prestwich: or make single not throw?

packages/sdk/src/NomadContext.ts
anna-carroll: we are working on all of the above FWIW

packages/sdk/src/NomadContext.ts
prestwich: cast it to the type we actually expect :)

prestwich:
Consider:

Should there be a `process` on `NomadMessage` that calles the `process` on the context?

packages/sdk/src/NomadContext.ts
ErinHales: Well, if the urls changed, we'd have to update them anyway.  This seems like a good clean way to do it

packages/sdk/src/NomadContext.ts
ErinHales: I don't think there's a use case for `getProof` other than processing, so I think it makes sense to leave as is for now.

packages/sdk/src/NomadContext.ts
ErinHales: Good call, thanks! I don't think getReplicaFor connects the signer, but I could be wrong

packages/sdk/src/messages/NomadMessage.ts
```diff
@@ -433,6 +434,10 @@ export class NomadMessage<T extends NomadContext> {
     return this.replica.confirmAt(newRoot);
   }
 
+  async process(): Promise<ContractTransaction> {
+    return this.process()
```
prestwich, line 438: this is just an infinite loop? needs to be like `this.context.process`

prestwich (changes requested):

packages/sdk/src/NomadContext.ts
```diff
@@ -1,12 +1,21 @@
...
 import { CoreContracts } from './CoreContracts';
+import { NomadMessage } from './messages/NomadMessage'
 
 export type Address = string;
+type MessageProof = {
+  message: BytesLike;
+  proof: {
+    leaf: BytesLike;
+    index: number;
+    path: [BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike, BytesLike];
```
prestwich, line 16: make a type alias for this

prestwich: future:
- import the type from the new accumulator npm package

prestwich (changes requested):

packages/sdk/src/NomadContext.ts
ErinHales: There was a way I used to do this type in the GUI that was really clean, been trying to find it :(

packages/sdk/src/NomadContext.ts
```diff
@@ -173,6 +189,43 @@ export class NomadContext extends MultiProvider<config.Domain> {
...
+   * @returns The Contract Transaction receipt
+   */
+  async process(message: NomadMessage<NomadContext>): Promise<ContractTransaction>{
+    const s3URL = `https://nomadxyz-${this.environment}-proofs.s3.us-west-2.amazonaws.com/`
+
+    const originNetwork = this.resolveDomainName(message.origin)
+    const destNetwork = this.resolveDomainName(message.destination)
+    const index = message.leafIndex.toString()
+    const s3Res = await fetch(`${s3URL}${originNetwork}_${index}`)
+    const data: MessageProof = await s3Res.json()
```
prestwich, line 213: what happens if the proof isn't in s3?

prestwich:
need to run prettier :)

prestwich:
oh wait you did already lol

packages/sdk/src/NomadContext.ts
```diff
@@ -174,6 +190,44 @@ export class NomadContext extends MultiProvider<config.Domain> {
...
+    const index = message.leafIndex.toString();
+    const s3Res = await fetch(`${s3URL}${originNetwork}_${index}`);
+    if (!s3Res) throw new Error('Not able to fetch proof');
+    const data: MessageProof = await s3Res.json();
+
+    // get replica contract
+    const replica = this.mustGetReplicaFor(originNetwork, destNetwork);
+
+    // get signer and connect replica
+    const signer = this.mustGetSigner(destNetwork);
```
prestwich, line 221: I think this is unnecessary? as `getReplica` is the underlying implementation here, and that does connect

https://github.com/nomad-xyz/monorepo/blob/f697b3649fc179dc3391fe1b0f68cb26062c6fad/packages/new-deploy/src/core/CoreContracts.ts#L127-L137

ErinHales: Cool! Yeah, that works, just tested it out

prestwich (approved):
LGTM

PR #224, opened 2022-04-20
refactor: store `CallBatch` on `DeployContext`

anna-carroll:
## Motivation

DeployContext was built with Object-Oriented pattern in mind, however, governance transactions (CallBatches) were returned in a functional manner. This made more flexible changes adding new governance transactions more difficult.

## Solution

Refactor to store the CallBatch on the DeployContext object.

## PR Checklist

- [ ] Added Tests
- [x] Updated Documentation

### DIFF
--- a/packages/new-deploy/scripts/deploy.ts
+++ b/packages/new-deploy/scripts/deploy.ts
@@ -28,24 +28,24 @@ async function run() {
     deployContext.registerSigner(network, signer);
     deployContext.overrides.set(network, OVERRIDES[network]);
   }
-  // run the deploy script
   const outputDir = './output';
   try {
-    const governanceBatch = await deployContext.deployAndRelinquish();
+    // run the deploy script
+    await deployContext.deployAndRelinquish();
 
+    // output the updated config & verification
     outputConfigAndVerification(outputDir, deployContext);
 
-    if (governanceBatch && !governanceBatch.isEmpty()) {
-      // build & write governance batch
-      await governanceBatch.build();
-      fs.writeFileSync(
-          `${outputDir}/governanceTransactions.json`,
-          JSON.stringify(governanceBatch, null, 2),
-      );
-    }
+    // output the governance transactions when the deploy succeeds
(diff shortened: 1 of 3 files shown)


packages/new-deploy/scripts/deploy.ts
```diff
@@ -28,24 +28,24 @@ async function run() {
...
-      await governanceBatch.build();
-      fs.writeFileSync(
-          `${outputDir}/governanceTransactions.json`,
-          JSON.stringify(governanceBatch, null, 2),
-      );
-    }
+    // output the governance transactions when the deploy succeeds
+    // (if the deploy script throws, the callBatch will be
+    // re-generated idempotently when the script ultimately succeeds)
+    await outputCallBatch(outputDir, deployContext);
```
kekonen, line 42: Could we use methods and standalone functions, because it feels like a step backwards towards the old deploy

packages/new-deploy/scripts/deploy.ts
```diff
@@ -28,24 +28,24 @@ async function run() {
...
 
     console.log(`DONE!`);
   } catch (e) {
+    // if the deploy script fails,
+    // output the updated config & verification
+    // then re-run the script to pick up where it left off
     outputConfigAndVerification(outputDir, deployContext);
```
kekonen, line 49: Same thing as the comment above (i forgot to add in the previous PR, but I think it is not too late)

kekonen (changes requested):
Basically these 2. I didn't add into the last PR the comment for a standalone function, but I think it makes sense to use standalone functions less, so that we will not come back to the old deploy which consisted almost only of standalone functions

packages/new-deploy/scripts/deploy.ts
anna-carroll: hmmm yeah, all of our code for reading / writing from local files is right now in a functional way rather than the Object-Oriented, but we could refactor to move it into the object perhaps

kekonen (approved):

PR #249, opened 2022-04-26
chore: remove old deploy, rename new-deploy

Imti:
## Motivation

Old deploy no longer being used, new-deploy is being used

## Solution

Remove deploy, rename new-deploy to deploy

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


packages/contracts-bridge/hardhat.config.ts
```diff
@@ -2,23 +2,24 @@ import "hardhat-gas-reporter";
 import "@nomiclabs/hardhat-etherscan";
 import "hardhat-packager";
 
-import { task } from "hardhat/config";
-import { verifyBridgeDeploy } from "@nomad-xyz/deploy/src/verification/verifyDeploy";
+// import { task } from "hardhat/config";
```
anna-carroll, line 5: you can just delete this!

packages/contracts-core/hardhat.config.ts
```diff
@@ -2,23 +2,24 @@ import "hardhat-gas-reporter";
 import "@nomiclabs/hardhat-etherscan";
 import "hardhat-packager";
 
-import { task } from "hardhat/config";
-import { verifyCoreDeploy } from "@nomad-xyz/deploy/src/verification/verifyDeploy";
+// import { task } from "hardhat/config";
```
anna-carroll, line 5: ditto, delete!

packages/local-environment/src/network.ts
```diff
@@ -5,6 +5,7 @@ import { Key } from ".";
 import { sleep } from "./utils";
 import { DockerizedActor } from "./actors";
 import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
+// TODO: the old deploy (what is currently published to npm) has been deprecated
```
anna-carroll, line 8: @kekonen can we merge this without these TODOs done, or do they need to be addressed?

anna-carroll (commented):
looks good!!

anna-carroll:
I think this is the only thing in the old deploy that's not in the new deploy which we use: https://github.com/nomad-xyz/monorepo/issues/258

but we can merge / delete old deploy without it

packages/local-environment/src/network.ts
Imti: local-environment seems to be out of date anyways. Would love to delete it in a separate PR

Imti: PR for deleting local-environment (https://github.com/nomad-xyz/monorepo/pull/266)

yourbuddyconner (approved):

PR #270, opened 2022-05-03
bug: allow max-balance send

ErinHales:
Bug where user can't send max amount

## Solution

Fix conditional statement to check if approve amount is greater than user's balance

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


ErinHales:
Just kidding.  Misread the function, it's correct as is

PR #280, opened 2022-05-12
add: prepare sends

ErinHales:
Expose the unsigned transaction, for allowing bridging using fireblocks wallet

## Motivation

Enable partners to bridge using fireblocks

## Solution

Add `prepareSend` and `prepareSendNative` that expose the unsigned tx

## PR Checklist

- [x] Added Tests (no additional code, just splits up existing functions)
- [x] Updated Documentation

### DIFF
--- a/packages/sdk-bridge/CHANGELOG.md
+++ b/packages/sdk-bridge/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ### Unreleased
 
+### 1.0.0-rc.14
+
+- add: prepareSend, prepareSendNative
+
 ### 1.0.0-rc.13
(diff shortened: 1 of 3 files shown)


prestwich:
is there any fireblocks ethers provider? 🤔 It seems like these should be
solved via a lower-level integration, rather than at the sdk

yourbuddyconner:
> is there any fireblocks ethers provider? 🤔 It seems like these should be solved via a lower-level integration, rather than at the sdk

Short answer, no, long answer, it's complicated.

yourbuddyconner (approved):

PR #289, opened 2022-05-17
fix: prevent messages rendered invalid during recovery from eventually being processed

prestwich:
## Motivation

Currently if a message is proven under a fraudulent root, it can eventually be processed even if the root is deleted from the confirmAt mapping by governance action. This means that the governance would need to clean each fraudulent message individually

## Solution

- instead of storing a messagestatus enum, prove stores the root that was calculated for the proof
- process checks that the root is acceptable again before dispatch

- for backwards compatibility `MessageStatus.Proven` is hardcoded as an acceptable root.

- we now allow re-proving a message under a new root. This ensures that VALID messages proved under an INVALID root can be processed after fraud recovery

- This SHOULD NOT break storage as we simply change the semantics of a mapping, rather than the storage layout
- This WILL require mainnet fork testing to ensure that it doesn't break storage in unexpected ways

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-core/contracts/Replica.sol
+++ b/packages/contracts-core/contracts/Replica.sol
@@ -24,18 +24,6 @@ contract Replica is Version0, NomadBase {
     using TypedMemView for bytes29;
     using Message for bytes29;
 
-    // ============ Enums ============
-
-    // Status of Message:
-    //   0 - None - message has not been proven or processed
-    //   1 - Proven - message inclusion proof has been validated
-    //   2 - Processed - message has been dispatched to recipient
-    enum MessageStatus {
-        None,
-        Proven,
-        Processed
-    }
-
     // ============ Public Storage ============
 
     // Domain of home chain
@@ -47,7 +35,7 @@ contract Replica is Version0, NomadBase {
     // Mapping of roots to allowable confirmation times
     mapping(bytes32 => uint256) public confirmAt;
     // Mapping of message leaves to MessageStatus
-    mapping(bytes32 => MessageStatus) public messages;
+    mapping(bytes32 => bytes32) public messages;
 
     // ============ Upgrade Gap ============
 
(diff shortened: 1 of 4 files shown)


packages/contracts-core/contracts/Replica.sol
```diff
@@ -52,8 +40,11 @@ contract Replica is Version0, NomadBase {
     uint8 private entered;
     // Mapping of roots to allowable confirmation times
     mapping(bytes32 => uint256) public confirmAt;
-    // Mapping of message leaves to MessageStatus
-    mapping(bytes32 => MessageStatus) public messages;
+    // Mapping of message leaves to their status
```
anna-carroll, line 43: shouuuuuld we add a helper function for off-chain actors.. 
```
function messageStatus(bytes32 _messageHash) external view returns (uint256 _status)
```

packages/sdk/src/messages/NomadMessage.ts
```diff
@@ -447,7 +448,12 @@ export class NomadMessage<T extends NomadContext> {
    * status of the message.
    */
   async replicaStatus(): Promise<ReplicaMessageStatus> {
-    return this.replica.messages(this.leaf);
+    const root = await this.replica.messages(this.leaf);
```
anna-carroll, line 451: will this still work before we perform the upgrade?

do we need
```
    if (root === ethers.constants.HashZero || root === 0) return ReplicaMessageStatus.None;
    if (root === `0x${'00'.repeat(31)}02` || root == 2)
      return ReplicaMessageStatus.Processed;
```

anna-carroll (changes requested):
what if a valid message is proven under an invalid root before that root is removed?

packages/sdk/src/messages/NomadMessage.ts
prestwich: fixed

packages/contracts-core/contracts/Replica.sol
```diff
@@ -23,17 +23,10 @@ contract Replica is Version0, NomadBase {
...
-    //   0 - None - message has not been proven or processed
-    //   1 - Proven - message inclusion proof has been validated
-    //   2 - Processed - message has been dispatched to recipient
-    enum MessageStatus {
-        None,
-        Proven,
-        Processed
-    }
+    // ============ Constants ============
+    bytes32 public constant LEGACY_STATUS_NONE = bytes32(0);
```
anna-carroll, line 27: this is unused

anna-carroll (changes requested):

packages/contracts-core/contracts/Replica.sol
prestwich: preserved for completeness

anna-carroll (approved):

PR #291, opened 2022-05-17
feature: instantiate SDK from static hosted config

prestwich:
Adds a static method to `NomadContext` that allows fetching a config from the confis repo by env name. 

## Motivation
Allow the SDK to update to the latest version of a config by default, without publishing a new configuration package version

## Solution

- add a `fetchConfig` and a `fetch` method to the NomadContext

## PR Checklist

- [ ] Added Tests
- [x] Updated Documentation
- [x] Updated CHANGELOG.md for the appropriate package

PR #292, opened 2022-05-17
Deployment scripts: quality of life improvement

xBalbinus:
## Motivation

Currently, configs for deployment are located on the rust repo, instead of in the same folder as the deploy script. On deploy, one (at least in my attempt) has to ferry over the dev configs. When asked, this was considered as a potential improvement so wanted to make a PR here to flag.

## Solution

Move the deployment configs to the config folder in `packages/deploy`

## PR Checklist

- [] Added Tests
- [] Updated Documentation


xBalbinus:
p.s. the deletion on .env.example is NOT intentional and will be fixed.

CONTRIBUTING.md
```diff
@@ -68,10 +68,19 @@ If needed, reference Foundry's [Troubleshooting Installation](https://github.com
...
 
 ```
-$ yarn
-$ yarn bootstrap
+$ yarn install
+$ yarn build
+$ yarn compile
 ```
 
+After that, go to the `configuration` folder found in the [rust repo](https://github.com/nomad-xyz/rust). Pull whichever config you want from there, and place that inside your config folder in `/packages/deploy/config`.
```
prestwich, line 76: rather than copying these from the configuration, the intended progression is via #291, where the latest config is loaded from a URL. The configs should not be committed multiple places. We want a single source of truth

Side note, the contributing doc is for regular contributors, whom we want to discourage from using the deploy script directly. The deploy script is primarily intended for use by the team and by power users performing governance actions. I'd avoid setting it up in `CONTRIBUTING.md`

xBalbinus: Thanks so much for the feedback. Closing this PR as proposed changes are unnecessary & addressed by #291

PR #293, opened 2022-05-18
Indexer Graphql and RPC rate limiting

kekonen:
## Solution
* adds secret ingredient which I cannot tell about at this moment
* added rpc rate limiting
* fixed a graphql query
* small fixes in the api

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


packages/indexer/src/core/db.ts
```diff
@@ -188,18 +188,34 @@ export class DB {
   }
 
   async getMessages(req: MsgRequest): Promise<NomadMessage[]> {
-    const take = req.size || 15;
-    const page = req.page || 1;
-    const skip = (page || -1) * take;
+    const take = req.size ? parseInt(req.size) : 15;
+    const page = req.page ? parseInt(req.page) - 1 : 0;
```
Imti, line 192: Are `req.size` and `req.page` strings here? Gotta be careful with `0` since it's truthy which can lead to a `-1` page value

packages/indexer/Dockerfile
```diff
@@ -10,8 +10,8 @@ COPY package.json ./package.json
 # COPY package-lock.json ./package-lock.json
 RUN yarn install
 
-RUN rm -rf node_modules
-RUN npm i
+# RUN rm -rf node_modules
```
Imti, line 13: We can remove these

packages/indexer/src/api/index.ts
```diff
@@ -10,6 +17,12 @@ import * as dotenv from 'dotenv';
 import Logger from 'bunyan';
 import promBundle from 'express-prom-bundle';
 
+import { resolvers } from '@generated/type-graphql';
+// import { buildSchema } from 'graphql';
```
Imti, line 21: Let's remove all the commented out code

packages/indexer/src/api/index.ts
```diff
@@ -251,18 +300,31 @@ export async function run(db: DB, logger: Logger) {
     '/tx',
     log,
     async (req: express.Request<{}, {}, {}, MsgRequest>, res) => {
-      const { size } = req.query;
+      const { size: sizeStr } = req.query;
 
-      if (size && size > 30) return fail(res, 403, 'maximum page size is 30');
+      try {
+        if (sizeStr && parseInt(sizeStr) > 30)
```
Imti, line 306: Why is the max size 30? Just curious

packages/indexer/src/core/indexer.ts
```diff
@@ -126,7 +126,9 @@ export class Indexer {
     this.domain = domain;
     this.sdk = sdk;
     this.orchestrator = orchestrator;
-    this.develop = false;
+    this.develop = !!(
```
Imti, line 129: Might be able to do `this.develop = process.env.NODE_ENV === 'development'` here. Also do we need staging specific logic as well or just for development?

packages/indexer/src/core/indexer.ts
```diff
@@ -263,20 +267,19 @@ export class Indexer {
...
-          const start = new Date().valueOf();
-          const r = await this.provider.getBlock(blockNumber);
-          this.orchestrator.metrics.observeRpcLatency(
-            RpcRequestMethod.GetBlock,
-            this.network,
-            new Date().valueOf() - start,
-          );
-          return r;
-        });
+        this.orchestrator.metrics.incRpcRequests(
```
Imti, line 270: Love that you added latency metrics 😻 

Might be able to make this block into a function that takes in arguments, just a thought.

packages/indexer/src/core/indexer.ts
```diff
@@ -623,13 +625,15 @@ export class Indexer {
...
-        tries += 1;
-        this.dummyTestEventsIntegrity();
-        passed = true;
-      } catch (e) {
-        this.logger.warn(`Dummy test not passed:`, e);
-        from -= batchSize;
+      if (this.develop) {
+        try {
+          tries += 1;
+          this.dummyTestEventsIntegrity();
```
Imti, line 631: Need to await this or nah?

packages/indexer/src/core/rateLimiting.ts
```diff
@@ -0,0 +1,177 @@
...
+  };
+}
+
+export function tryLimitStrategyByName(
+  name: string,
+): CupsLimiterStrategy | RPPLimiterStrategy | undefined {
+  const _name = name.toLowerCase();
+  switch (_name) {
+    case 'moonbeam':
+      return strategyFromRPP(3.9); // 1.9
```
Imti, line 124: Where do these numbers come from?

packages/indexer/src/core/utils.ts
```diff
@@ -132,6 +132,37 @@ export function createLogger(
   });
 }
 
+function formatEVM(s: string) {
```
Imti, line 135: What does this do? Might be good to include an example as a comment right above this function

Imti (commented):
Left some comments, let me know if you need any help

Imti:
Can we add what changed to the changelog as well and bump the package version?

packages/indexer/src/core/db.ts
kekonen: the page value is tested in the next line to be not less than 0
`if (page < 0) throw new Error(`Page is less than a 0`);`
Though I will go with first page start with 0 and not with 1

packages/indexer/Dockerfile
kekonen: Agree

packages/indexer/src/api/index.ts
kekonen: the day has come...

packages/indexer/src/api/index.ts
kekonen: to prevent system from quering all of the messages at once. Though it is not so useful while we don't have rate limiting and other stuff

packages/indexer/src/core/indexer.ts
kekonen: yes, thanks! :)

packages/indexer/src/core/indexer.ts
kekonen: absolutely. but lets do later, priority now is to publish it as a stable thing asap

packages/indexer/src/core/indexer.ts
kekonen: thanks! yeah

packages/indexer/src/core/rateLimiting.ts
kekonen: good question! They are coming from the calculation of the relative weighted PRC requests per second. For example, [alchemy allows us to do 330 compute units (CU) per second](https://docs.alchemy.com/alchemy/documentation/throughput), where every method call has a fixed CU. ﻿﻿eth_blockNumber costs 10 CUs, eth_getLogs is 75. If we make every second 10 requests by 10 CUs and 10 by 75, then in total we make 850 CUs per second. In reality this number of allowed CU's/sec is waaay lower. (Also the approximate CU's are calculated for Alchemy, but we use the same numbers assuming that other RPC have somewhat similar mechanisms)

So having this knowledge I calculated the approx CU/sec for every RPC we run using [Graphana data](https://nomadxyz.grafana.net/d/hxT-q6-7z/indexer-production?orgId=1&var-network=ethereum&var-environment=production&viewPanel=42) and [spread sheets](https://docs.google.com/spreadsheets/d/16vicOJwCqMHHWZ1pmvw8DQYEGXTNs1uXQFrMszZ8694/edit#gid=0). The numbers I've got are in the // comment, and I fine-tuned them a bit, so u see 3.9 as the tuned number, which works alright, because alchemy works a bit differently to others. We should make this fine-tuning automatically using RPC tester which is in @yourbuddyconner 's plans for long time.

packages/indexer/src/core/utils.ts
kekonen: Actually, all of what's in there must have comments and the docs, but that comes right after this PR I believe.
This particular function is producing the evm's address formatted using [checksum](https://coincodex.com/article/2078/ethereum-address-checksum-explained/)

Imti (approved):
🏆

Imti:
Looks like `yarn install` is failing. Did we commit the yarn.lock file?

PR #298, opened 2022-05-19
[QSP-8, QSP-10, QSP-11] deprecate fast liquidity send path

anna-carroll:
## Motivation
Fast liquidity feature was mostly a proof of concept, which was not in use in production.

## Solution
We are choosing to deprecate it.

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-bridge/contracts/BridgeMessage.sol
+++ b/packages/contracts-bridge/contracts/BridgeMessage.sol
@@ -134,19 +134,17 @@ library BridgeMessage {
      * @notice Formats Transfer
      * @param _to The recipient address as bytes32
      * @param _amnt The transfer amount
-     * @param _enableFast True to format FastTransfer, False to format regular Transfer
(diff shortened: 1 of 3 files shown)


anna-carroll:
initially tried having two `send` functions - one without `enableFast` param, and one with - but this causes a problem when compiling the contracts to typescript. it would mean that users of the typescript sdk could no longer call the contract `bridgeRouter.send(...`, but instead would have to cal `bridgeRouter."send(address,uint256,uint32,bytes32)"(...`

after discussing with typescript engineers, i think the only way to fix this is to not have the function name `send` overloaded:
1. just leave the old `send` function, with the boolean where `enableFast` was, but just always ignore that param 
2. leave the old `send` function, add the cleaned up function under a different name like `transfer` 

seems like option 1 is probably the cleaner choice

prestwich (approved):

PR #300, opened 2022-05-20
[QSP-2] remove FAILED state from Replica

anna-carroll:
## Motivation
The xApp is in control of determining when to block a Replica contract. The security layer for blocking fraud is at the xAppConnectionManager, in the hands of the xApp.
Additionally, we are moving in a direction where the concept of a `doubleUpdate` - and therefore, proving Fraud on the Replica - are irrelevant.

## Solution
Remove `doubleUpdate` proofs and ability to transition to FAILED state from the Replica.

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-core/contracts/Replica.sol
+++ b/packages/contracts-core/contracts/Replica.sol
@@ -64,8 +64,9 @@ contract Replica is Version0, NomadBase {
 
     /**
      * @notice Emitted when message is processed
-     * @param messageHash Hash of message that failed to process
-     * @param success TRUE if the call was executed successfully, FALSE if the call reverted
+     * @param messageHash Hash of message that was processed
+     * @param success TRUE if the call was executed successfully,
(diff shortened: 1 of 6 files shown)


packages/contracts-core/contracts/Replica.sol
```diff
@@ -64,8 +64,9 @@ contract Replica is Version0, NomadBase {
 
     /**
      * @notice Emitted when message is processed
-     * @param messageHash Hash of message that failed to process
-     * @param success TRUE if the call was executed successfully, FALSE if the call reverted
+     * @param messageHash Hash of message that was processed
+     * @param success TRUE if the call was executed successfully,
+     * FALSE if the call reverted or threw
```
prestwich, line 69: `reverted` and `threw` are the same

prestwich (approved):

PR #303, opened 2022-05-20
[QSP-9, QSP-20] add bounds to optimisticSeconds

anna-carroll:
## Motivation
We wish to add some reasonable input validation to the optimisticSeconds field

## Solution
Ensure optimisticSeconds is greater than 25 minutes & less than 1 year. 

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-core/contracts/Replica.sol
+++ b/packages/contracts-core/contracts/Replica.sol
@@ -134,8 +134,7 @@ contract Replica is Version0, NomadBase {
         committedRoot = _committedRoot;
         // pre-approve the committed root.
         confirmAt[_committedRoot] = 1;
-        optimisticSeconds = _optimisticSeconds;
-        emit SetOptimisticTimeout(_optimisticSeconds);
+        _setOptimisticTimeout(_optimisticSeconds);
     }
 
     // ============ External Functions ============
@@ -274,8 +273,7 @@ contract Replica is Version0, NomadBase {
         external
         onlyOwner
     {
-        optimisticSeconds = _optimisticSeconds;
-        emit SetOptimisticTimeout(_optimisticSeconds);
+        _setOptimisticTimeout(_optimisticSeconds);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


anna-carroll:
realized while making this PR that, ofc, we set `optimisticTimeout` to 10 seconds in our development testnet environment 😭

odyslam:
Although I am generally against input validation for gas optimization reasons, I think this codepth is rare enough to allow for input validation

anna-carroll:
> rare enough to allow for input validation

yes, but for our development testnet environment, we want to use a value for `optimisticSeconds` which is not within the bounds of reason for production (10 seconds for dev environment)

packages/contracts-core/contracts/Replica.sol
```diff
@@ -345,6 +343,22 @@ contract Replica is Version0, NomadBase {
...
     // ============ Internal Functions ============
 
+    /**
+     * @notice Set optimistic timeout period for new roots
+     * @dev Called by owner (Governance) or at initialization
+     * @param _optimisticSeconds New optimistic timeout period
+     */
+    function _setOptimisticTimeout(uint256 _optimisticSeconds) internal {
+        // ensure the optimistic timeout is at least 25 minutes
+        require(_optimisticSeconds >= 1500, "optimistic timeout too low");
```
prestwich, line 353: what about testing environments?

anna-carroll:
we don't want to address this for test environments

prestwich:
We want to address this, but with no minimum bound

packages/contracts-core/contracts/Replica.sol
```diff
@@ -359,6 +357,25 @@ contract Replica is Version0, NomadBase {
...
+     * @notice Set optimistic timeout period for new roots
+     * @dev Called by owner (Governance) or at initialization
+     * @param _optimisticSeconds New optimistic timeout period
+     */
+    function _setOptimisticTimeout(uint256 _optimisticSeconds) internal {
+        // This allows us to initialize the value to be very low in test envs,
+        // but does not allow governance action to lower a production env below
+        // the safe value
+        uint256 _current = optimisticSeconds;
+        if (_current != 0 && _current > 1500) require(_optimisticSeconds >= 1500, "optimistic timeout too low");
```
prestwich, line 370: @anna-carroll please review this line carefully

prestwich (approved):
approved, provided you thoroughly check my if/require

packages/contracts-core/contracts/Replica.sol
anna-carroll: hmm prevents us from updating `optimisticSeconds` in test envs. should be acceptable tho.. let me think a sec

prestwich: it shouldn't prevent us from updating in test envs. It should be a ratchet. I.e. once above 1500, never below 1509

anna-carroll: what if 

if current is zero, allow anything 
if current is less than 1500, require it's greater than current
if current is geq 1500, require it's geq 1500

anna-carroll: yep, okay. good

anna-carroll (commented):
good.

anna-carroll:
if current = 0 (never set before), allow anything
if current < 1500 (testnet environments), allow anything
if current > 1500 (prod), require GEQ 1500

PR #305, opened 2022-05-24
docs: address QS report docs

prestwich:
Address doc issues from QS report

## PR Checklist

- [n/a] Added Tests
- [x] Updated Documentation
- [n/a (non-functional)] Updated CHANGELOG.md for the appropriate package

### DIFF
--- a/packages/contracts-bridge/contracts/TokenRegistry.sol
+++ b/packages/contracts-bridge/contracts/TokenRegistry.sol
@@ -152,7 +152,7 @@ contract TokenRegistry is Initializable, XAppConnectionClient, ITokenRegistry {
      * The custom representation MUST be trusted, and MUST allow the router to
      * both mint AND burn tokens at will.
      * @param _domain the domain of the canonical Token to enroll
-     * @param _id the bytes32 ID pf tje canonical of the Token to enroll
+     * @param _id the bytes32 ID of the canonical of the Token to enroll
      * @param _custom the address of the custom implementation to use.
      */
     function enrollCustom(
@@ -341,7 +341,9 @@ contract TokenRegistry is Initializable, XAppConnectionClient, ITokenRegistry {
     {
         // deploy and initialize the token contract
         _token = address(new UpgradeBeaconProxy(tokenBeacon, ""));
-        // initialize the token separately from the
+        // Initialize the token. Normally we initialize during proxy deployment
+        // but due to a proxy quirk, it's clearer to do it explicitly when
+        // calling an initializer with no arguments.
         IBridgeToken(_token).initialize();
         // set the default token name & symbol
         (string memory _name, string memory _symbol) = _defaultDetails(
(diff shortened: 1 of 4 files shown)


packages/contracts-core/contracts/Replica.sol
```diff
@@ -109,6 +109,19 @@ contract Replica is Version0, NomadBase {
 
     // ============ Initializer ============
 
+    /**
+     * @notice Iniitalize the replica
```
odyslam, line 113:
```suggestion
     * @notice Initialize the replica
```

packages/contracts-bridge/contracts/TokenRegistry.sol
```diff
@@ -341,7 +341,9 @@ contract TokenRegistry is Initializable, XAppConnectionClient, ITokenRegistry {
     {
         // deploy and initialize the token contract
         _token = address(new UpgradeBeaconProxy(tokenBeacon, ""));
-        // initialize the token separately from the
+        // Initialize the token. Normally we initialize during proxy deployment
+        // but due to a proxy quirk, it's clearer to do it explicitly when
```
odyslam, line 345: what's a "proxy quirk"? Perhaps we can add a sentence or point the user somewhere

packages/contracts-core/contracts/Home.sol
```diff
@@ -51,11 +51,13 @@ contract Home is Version0, QueueManager, MerkleTreeManager, NomadBase {
...
      * @notice Emitted when a new message is dispatched via Nomad
+     * @param messageHash Hash of message; the leaf inserted to the Merkle tree
+     *        for the message
      * @param leafIndex Index of message's leaf in merkle tree
      * @param destinationAndNonce Destination and destination-specific
-     * nonce combined in single field ((destination << 32) & nonce)
-     * @param messageHash Hash of message; the leaf inserted to the Merkle tree for the message
-     * @param committedRoot the latest notarized root submitted in the last signed Update
+     *        nonce combined in single field ((destination << 32) & nonce)
+     * @param committedRoot the latest notarized root submitted in the last
```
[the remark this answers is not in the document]
prestwich: oh I prefer hanging indents to match the first meaningful word of the previous line. i.e. make it visually obvious it's a continuation line

packages/contracts-bridge/contracts/TokenRegistry.sol
prestwich: the proxy quirk is that if the initdata is length 0, initialize is not called

prestwich: don't think we need to specify more

anna-carroll: for the record i think it makes more sense for a proxy to not call initialize if there's no initialize calldata passed in. what if the implementation doesn't *have* an initialize function?

anna-carroll: so this should really saw "due to awesome proxy design"

anna-carroll (approved):

PR #310, opened 2022-05-25
Deploy checks collect errors in place of throwing

kekonen:
## Motivation
We don't want to throw deploy checks during the actual deploy, but rather collect them all into a list and if there are any throw that list

## Solution
The solution could be found in the Motivation section

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/deploy/src/core/CoreContracts.ts
+++ b/packages/deploy/src/core/CoreContracts.ts
@@ -16,7 +16,7 @@ import Contracts from '../Contracts';
 import { DeployContext } from '../DeployContext';
 
 import { expect } from 'chai';
-import { log, assertBeaconProxy } from '../utils';
+import { CheckList, log } from '../utils';
 
 export abstract class AbstractCoreDeploy<T> extends Contracts<T> {
   // Placeholder for future multi-VM abstraction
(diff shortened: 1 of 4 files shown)


packages/deploy/src/core/CoreContracts.ts
```diff
@@ -915,36 +955,54 @@ export default class EvmCoreDeploy extends AbstractCoreDeploy<config.EvmCoreCont
...
-        expect(utils.equalIds(beacon, replicaBeacon));
-      });
+    checklist.equalIds(
+      beaconOwner,
+      this.governanceRouter.address,
+      'Governance router address',
+    );
+
+    if (remoteDomains.length > 0 && replicas) {
+      if (replicas) {
```
Imti, line 982: We can remove the `if (replicas)`

packages/deploy/src/core/CoreContracts.ts
```diff
@@ -915,36 +955,54 @@ export default class EvmCoreDeploy extends AbstractCoreDeploy<config.EvmCoreCont
...
+            replicaImpl,
+            "Replica's implementation",
+          );
+          checklist.equalIds(beacon, replicaBeacon, "Replica's beacon");
+        });
+      }
+    }
+
+    if (checklist.hasErrors()) {
+      throw checklist;
```
Imti, line 1003: When it throws, do you see the full `ok` and `errors` arrays?

Imti (approved):
🚀

packages/deploy/src/core/CoreContracts.ts
kekonen: it throws or returns the same object. `ok` and `error` are visible both ways because we always want to know what went good, what went bad

packages/deploy/src/core/CoreContracts.ts
kekonen: good call

PR #312, opened 2022-05-26
deploy from tagged contract version

anna-carroll:
## Motivation
Deploy repo was deploying from workspace version of the contracts, not tagged npm version. This didn't affect the source code deploys as we had frozen the contract code, but needs to be changed nevertheless. 

## Solution
Update deploy package to use tagged contract versions rather than workspace.

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/deploy/CHANGELOG.md
+++ b/packages/deploy/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Unreleased
 
+- deploy from tagged contract versions (1.0.0-rc.2)
 - feature: deploy script doesn't throw, but collect errors
 - chore: bump configuration to rc.23
 - fix: add goerli overrides to overrides.json
--- a/packages/deploy/package.json
+++ b/packages/deploy/package.json
@@ -22,9 +22,9 @@
   },
   "dependencies": {
     "@nomad-xyz/configuration": "^0.1.0-rc.23",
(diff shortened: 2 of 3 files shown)


Imti:
When do we need to bump the contract versions in deploy? For every time the contracts change or whenever we wanna deploy?

anna-carroll:
> When do we need to bump the contract versions in deploy? For every time the contracts change or whenever we wanna deploy?

we need to bump the contract versions when we release a new tagged version of the contracts, which will happen each time we perform a protocol sprint + audit + on-chain upgrade. this will be approx every 6-8 weeks.

odyslam:
Instead of hardcoding it in the package.json, could the deployment script `fetch` the latest npm versions from the npm registry?

prestwich:
putting it in the package.json will also cause problems when we want to deploy test contracts that aren't tagged on EVM. e.g. testing in local env?

running into rough edges on to my understanding of JS packaging

prestwich:
@anna-carroll because we may need to run deployments from commits older than this, you'll need to add "check the version" to your list of manual deploy pre-flight checks 😓

anna-carroll:
Okay I have created a ticket where we should discuss this: https://github.com/nomad-xyz/monorepo/issues/325

Closing PR until we have aligned on an approach

PR #316, opened 2022-05-27
Revert PR #299 (QSP-4)

odyslam:
We revert the PR, as we weren't happy with the mitigation. 

We intend to remove root ordering in the near future, but due to the complexity of the endeavor (in both designing a solution and upgrading the system), we will PR the removal at a later time.


anna-carroll (approved):
LlGTM

PR #322, opened 2022-05-27
Metric blocks to target

kekonen:
## Motivation
We want to know how many blocks are to tip for each of indexers

## Solution
Added a metric into prometheus

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/indexer/src/api/index.ts
+++ b/packages/indexer/src/api/index.ts
@@ -1,6 +1,6 @@
 import 'reflect-metadata';
 
-import express from 'express';
+import express, { Response } from 'express';
 import cors from 'cors';
 import { ApolloServer } from 'apollo-server-express';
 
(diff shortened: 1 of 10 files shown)


packages/indexer/src/api/index.ts
```diff
@@ -67,14 +70,36 @@ export async function run(db: DB, logger: Logger) {
     next();
   };
 
+  // Kludge. I don't know how to prevent promBundle from exposing metricsPath
+  const metricsPath = '/' + randomString(20); // '/metrics'
```
Imti, line 74: Is there a way to put the metrics path on a separate port so that it's no exposed?

Imti (approved):
lfg 🚢

packages/indexer/src/api/index.ts
Imti: This is something we can improve as we go along

PR #327, opened 2022-06-01
Keymaster v2

kekonen:
## Motivation

We needed keymaster v2, which would be robust, not python, can work with KMS signers.

## Solution

Here we go... Keymaster v2. 
Works as very close almost drop-in replacement with slight changes:
* config is similar, but allows different types of signers
* metrics names now follow 'nomad_keymaster_*' mask.
* metrics record transfers from banks to agents
* upper-thresholds now being calculated from the lower-threshold
* watchers have configurable separate thresholds

## PR Checklist
- [x] Have been tested in dev and stage
- [ ] Added Tests
- [ ] Updated Documentation


packages/keymaster/src/config.ts
```diff
@@ -0,0 +1,23 @@
...
+  relayer: string;
+  processor: string;
+}
+
+export interface INetwork {
+  name: string;
+  endpoint: string;
+  replicas: string[];
+  bank: string | AwsKmsSignerCredentials;
+  treshold: ethers.BigNumber;
```
arnaud036, line 17: typo: threshold

packages/keymaster/src/account.ts
```diff
@@ -0,0 +1,702 @@
...
+    ctx: Context,
+    options?: OptionalNetworkArgs
+  ) {
+    super(name, ctx);
+
+    let realAddress;
+    let threshold: ethers.BigNumberish;
+    if (typeof address === "string") {
+      realAddress = address;
+      threshold = options?.threshold || eth(1.337);
```
Imti, line 61: How did we get the default threshold?

packages/keymaster/src/kms/util/aws-kms-utils.ts
```diff
@@ -0,0 +1,122 @@
...
+  const EthAddr = `0x${address.slice(-40)}`; // take last 20 bytes as ethereum adress
+  return EthAddr;
+}
+
+export function findEthereumSig(signature: Buffer) {
+  const decoded = EcdsaSigAsnParse.decode(signature, "der");
+  const { r, s } = decoded;
+
+  const secp256k1N = new BN(
+    "fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141",
```
Imti, line 73: Where does this string come from?

packages/keymaster/tsconfig.json
```diff
@@ -0,0 +1,106 @@
+{
```
Imti, line 1: This should extend from `tsconfig.package.json`

packages/keymaster/config/keymaster.json
```diff
@@ -0,0 +1,66 @@
...
+    "networks": {
+        "goerli": {
+            "name": "goerli",
+            "replicas": ["rinkeby", "evmostestnet", "kovan"],
+            "endpoint": "https://goerli.infura.io/v3/5f65182f06a14f5182a2e9af8d8fff33",
+            "bank": {
+                "region": "us-west-2",
+                "keyId": "455a9119-5349-4288-a73c-08ae8d9569c4"
+              },
+            "treshold": "500000000000000000",
```
Imti, line 11: Can we change this from `treshold` -> `threshold`?

Imti (approved):
🔑

packages/keymaster/src/kms/util/aws-kms-utils.ts
kekonen: this code is coming from [this repo](https://github.com/rjchow/ethers-aws-kms-signer). Right below this line on the line 75 there is a comment which states that this is "max value on the curve" (meaning ECDSA). Googling it to doublecheck brought me to [this forum](https://bitcointalk.org/index.php?topic=175389.0), which confirms that

packages/keymaster/config/keymaster.json
kekonen: done

packages/keymaster/src/account.ts
kekonen: set it to 1, which is somewhat default

packages/keymaster/src/utils.ts
```diff
@@ -185,7 +185,7 @@ function reviver(key: any, value: any) {
 
 export function readConfig(l: string): KeymasterConfig {
   let s = fs.readFileSync(l, "utf8");
-  s = s.replaceAll("treshold", "threshold"); // there was a typo, so we want to make use the config doesn't have it
+  s = s.replace(/treshold/g, "threshold"); // there was a typo, so we want to make use the config doesn't have it
```
Imti, line 188: We can safely remove this line now

kekonen: done

PR #334, opened 2022-06-03
fix: try send/sendNative with callStatic

ErinHales:
Test if a transaction with succeed/fail before trying to send.

## Motivation

Don't spend gas if we know it's going to fail

## Solution

Implement `callStatic` call

## PR Checklist

- [x] Added Tests (tested locally)
- [x] Updated Documentation


packages/sdk-bridge/src/BridgeContext.ts
```diff
@@ -279,17 +279,27 @@ export class BridgeContext extends NomadContext {
       await tx.wait();
     }
 
-    const tx = await fromBridge.bridgeRouter.populateTransaction.send(
+    if (!overrides.gasLimit) {
+      overrides.gasLimit = BigNumber.from(350000)
```
prestwich, line 283: magic number to named constant?

anna-carroll (approved):
very nice :,)

packages/sdk-bridge/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### 1.0.0-rc.19
 
 - fix: numeric fault overflow when supporting infinite token approval
+- feat: try send/sendNative and process calls using callStatic
```
prestwich, line 8: shouldn't this be in unreleased?

prestwich (changes requested):

packages/sdk-bridge/src/BridgeContext.ts
```diff
@@ -279,17 +279,27 @@ export class BridgeContext extends NomadContext {
       await tx.wait();
     }
 
-    const tx = await fromBridge.bridgeRouter.populateTransaction.send(
+    if (!overrides.gasLimit) {
+      overrides.gasLimit = BigNumber.from(350000);
```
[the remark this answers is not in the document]
ErinHales: What do you mean by magic number?

You just want it like this?
```
const gasLimit = BigNumber.from(350000);
overrides.gasLimit = gasLimit
```

packages/sdk-bridge/CHANGELOG.md
ErinHales: That package was about to be released but I asked to get this in at the same time

prestwich (approved):

PR #338, opened 2022-06-07
feat: Export version from TS package

prestwich:
Allow downstream users to identify the contract version via the contract TS packages

## Motivation

Future work may want to know whether it's using a compatible contract version. E.g. deploy script running updates may wish to know that it's on a later version than the previous contract version

## Solution

export `version` at the root level of each contract package

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


Imti (approved):
🌋

PR #347, opened 2022-06-16
feat: add config file for all dotenv environment variables

Imti:
## Motivation

Noticed we were using environment variables with defaults in a few different places

## Solution

A config file to hold all config variables with the default values in there for other files to import.

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/indexer/src/main.ts
+++ b/packages/indexer/src/main.ts
@@ -1,5 +1,4 @@
 import * as core from './core';
-import * as api from './api';
 import { DB } from './core/db';
 import { createLogger } from './core/utils';
 import { IndexerCollector } from './core/metrics';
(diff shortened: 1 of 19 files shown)


packages/indexer/src/main.ts
```diff
@@ -1,19 +1,18 @@
...
-export type NomadEnvironment = 'development' | 'staging' | 'production';
-export type Program = 'api' | 'core';
-
-const environment = process.env.ENVIRONMENT as NomadEnvironment;
-const configOverrideLocation = process.env.CONFIG_OVERRIDE_LOCATION;
-const program = process.env.PROGRAM as Program;
-const logLevel = (process.env.LOG_LEVEL || 'debug') as BunyanLevel;
-const metricsPort = parseInt(process.env.METRICS_PORT || '9090');
+import {
+  program,
```
barbaraliau, line 9: program vs Program?

kekonen: Program is the enum that contains multiple variations of what program (small p) could be

kekonen (approved):
lgtm

PR #363, opened 2022-06-23
fix: re-add metrics server

Imti:
## Motivation

Metrics server was not being started. Pulled the logic out to its own function, but forgot to call that function.

## Solution

Call the function to start the metrics server

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


Imti:
Not needed anymore after merging this https://github.com/nomad-xyz/monorepo/pull/366

PR #369, opened 2022-06-28
feat: add back script to calculate domain number

odyslam:

### DIFF
--- a/packages/deploy/scripts/domain.ts
+++ b/packages/deploy/scripts/domain.ts
@@ -1,0 +1,12 @@
+import { utils } from '@nomad-xyz/multi-provider';
+(async () => {
+  const args = process.argv.slice(2);
+  const network = args[0];
+  if (!network) {
+    console.error('Error: You need to pass a network as command line argument');
+    process.exit(1);
+  }
+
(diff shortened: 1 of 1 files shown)


packages/deploy/scripts/domain.ts
```diff
@@ -0,0 +1,12 @@
...
+(async () => {
+  const args = process.argv.slice(2);
+  const network = args[0];
+  if (!network) {
+    console.error('Error: You need to pass a network as command line argument');
+    process.exit(1);
+  }
+
+  const hexBytes = utils.getDomainFromString(network);
+  console.log(`${network} interpreted as hex number: ${hexBytes}`);
```
anna-carroll, line 11: when we add it to the config it needs to be a decimal number, so could we possibly also print the decimal version + the hex version?

prestwich: this is the decimal version. variable name and log should be changed to reflect that

https://github.com/nomad-xyz/monorepo/blob/main/packages/multi-provider/src/utils.ts#L48-L60

prestwich (approved):

PR #370, opened 2022-06-29
refactor: change message status from enum to tagged union

prestwich:
## Motivation

New protocol versions return the root a message was proven under. We should capture and expose this in the TS SDK

## Solution

Change MessageStatus from an enum to a union with discrimanator.

This is an API-breaking change (major version bump) however, given that we are in RCs this is still a candidate for the current non-RC release (i.e. we don't need to bump to 3.0.0 as 2.0.0 is not finalized)

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/sdk/src/messages/NomadMessage.ts
+++ b/packages/sdk/src/messages/NomadMessage.ts
@@ -494,7 +494,7 @@ export class NomadMessage<T extends NomadContext> {
    */
   async delivered(): Promise<boolean> {
     const { status } = await this.replicaStatus();
-    return status === 'processed';
+    return status === ReplicaStatusNames.Processed;
   }
 
   /**
--- a/packages/sdk/src/messages/NomadMessage.ts
+++ b/packages/sdk/src/messages/NomadMessage.ts
@@ -48,11 +48,23 @@ export enum MessageStatus {
   Processed = 3,
 }
 
-export enum ReplicaMessageStatus {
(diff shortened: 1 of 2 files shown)


packages/sdk/src/messages/NomadMessage.ts
```diff
@@ -450,13 +462,22 @@ export class NomadMessage<T extends NomadContext> {
   async replicaStatus(): Promise<ReplicaMessageStatus> {
     // backwards compatibility. Older replica versions returned a number,
     // newer versions return a hash
-    const root = (await this.replica.messages(this.leaf)) as string | number;
+    let root: string | number = await this.replica.messages(this.leaf);
+    root = root as string | number;
```
anna-carroll, line 466: is this necessary given the type is also defined above?

anna-carroll (commented):
i feel eh about this because it makes the code uglier, requires an API version change, and imo doesn't provide much end-user value 

but code lgtm

packages/sdk/src/messages/NomadMessage.ts
prestwich: yes, without this, it binds root's type as `string` 🙄

prestwich:
turning on auto-merge pending approval from indexer folks :)

ErinHales (approved):

PR #371, opened 2022-07-01
feat: add connext reconciliation + sending

LayneHaber:
## Motivation

Allow Connext to be integrated directly into the `BridgeRouter` instead of migration.

## Solution

- Add a reference to `Connext` to the `BridgeRouter`
- Add a `bytes32 _externalId` to the `BridgeMessage`. This identifier will only be set for Connext transfers
- Add a new `Type` for `BridgeMessage` of `ConnextTransfer`, which will include a non-zero identifier, and empty recipient
- Add `xsend` method to be used by Connext. This method acts exactly like `send`, but will create a `ConnextTransfer` message (recipient will be destination domain Connext contract, which is stored on the `BridgeRouter` so is empty within the message and there is a 32 byte identifier).
- On `handle` if it is a `ConnextTransfer` message, it will mint funds to the stored Connext contract and call `reconcile` with the external id and other token information

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-bridge/contracts/BridgeMessage.sol
+++ b/packages/contracts-bridge/contracts/BridgeMessage.sol
@@ -38,7 +38,8 @@ library BridgeMessage {
 
     uint256 private constant TOKEN_ID_LEN = 36; // 4 bytes domain + 32 bytes id
     uint256 private constant IDENTIFIER_LEN = 1;
-    uint256 private constant TRANSFER_LEN = 129; // 1 byte identifier + 32 bytes recipient + 32 bytes amount + 32 bytes detailsHash + 32 bytes externalId
+    uint256 private constant TRANSFER_LEN = 97; // 1 byte identifier + 32 bytes recipient + 32 bytes amount + 32 bytes detailsHash
+    uint256 private constant CONNEXT_TRANSFER_LEN = 97; // 1 byte identifier + 32 bytes externalId + 32 bytes amount + 32 bytes detailsHash
 
     // ============ Modifiers ============
 
@@ -60,7 +61,7 @@ library BridgeMessage {
      * @return TRUE if action is valid
      */
     function isValidAction(bytes29 _action) internal pure returns (bool) {
-        return isTransfer(_action) || isFastTransfer(_action);
+        return isTransfer(_action) || isFastTransfer(_action) || isConnextTransfer(_action);
     }
 
     /**
@@ -68,6 +69,8 @@ library BridgeMessage {
      * @param _view The bytes string
      * @return TRUE if message is valid
      */
+     // NOTE: this still works with CONNEXT_TRANSFER only because the `bytes32 recipient` is replaced with the `bytes32 externalId`
+     // should it be more explicit by adding an `||` condition on the _len?
     function isValidMessageLength(bytes29 _view) internal pure returns (bool) {
         uint256 _len = _view.len();
         return _len == TOKEN_ID_LEN + TRANSFER_LEN;
@@ -146,17 +149,29 @@ library BridgeMessage {
      * @param _to The recipient address as bytes32
      * @param _amnt The transfer amount
      * @param _detailsHash The hash of the token name, symbol, and decimals
-     * @param _externalId The external identifier of the transfer
      * @return
      */
     function formatTransfer(
         bytes32 _to,
         uint256 _amnt,
-        bytes32 _detailsHash,
-        bytes32 _externalId
+        bytes32 _detailsHash
     ) internal pure returns (bytes29) {
-        Types transferType = _externalId == bytes32(0) ? Types.Transfer : Types.ConnextTransfer;
-        return abi.encodePacked(transferType, _to, _amnt, _detailsHash, _externalId).ref(0).castTo(uint40(transferType));
+        return abi.encodePacked(Types.Transfer, _to, _amnt, _detailsHash).ref(0).castTo(uint40(Types.Transfer));
+    }
+
+    /**
+     * @notice Formats Connext Transfer
+     * @param _externalId The external identifier of the transfer
+     * @param _amnt The transfer amount
+     * @param _detailsHash The hash of the token name, symbol, and decimals
+     * @return
+     */
+    function formatConnextTransfer(
+        bytes32 _externalId,
+        uint256 _amnt,
+        bytes32 _detailsHash
+    ) internal pure returns (bytes29) {
+        return abi.encodePacked(Types.ConnextTransfer, _externalId, _amnt, _detailsHash).ref(0).castTo(uint40(Types.ConnextTransfer));
     }
 
     /**
@@ -306,6 +321,8 @@ library BridgeMessage {
      * @param _transferAction The message
      * @return The recipient address as bytes32
      */
+     // NOTE: If used on a Types.ConnextTransfer then this will return the external id. Should there
+     // be strict type assertion here as well? (this should be valid for Transfer or FastTransfer types)
     function recipient(bytes29 _transferAction)
         internal
         pure
@@ -354,13 +371,14 @@ library BridgeMessage {
     }
 
     /**
-     * @notice Retrieves the externalId from a Transfer
+     * @notice Retrieves the externalId from a Connext Transfer
+     * @dev MUST be used on the ConnextTransfer type, otherwise will revert
      * @param _transferAction The message
      * @return The external id
      */
-    function externalId(bytes29 _transferAction) internal pure returns (bytes32) {
+    function externalId(bytes29 _transferAction) internal pure typeAssert(_transferAction, Types.ConnextTransfer) returns (bytes32) {
         // before = 1 byte identifier
-        return _transferAction.index(97, 32);
+        return _transferAction.index(1, 32);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


packages/contracts-bridge/contracts/BridgeMessage.sol
```diff
@@ -37,7 +38,7 @@ library BridgeMessage {
 
     uint256 private constant TOKEN_ID_LEN = 36; // 4 bytes domain + 32 bytes id
     uint256 private constant IDENTIFIER_LEN = 1;
-    uint256 private constant TRANSFER_LEN = 97; // 1 byte identifier + 32 bytes recipient + 32 bytes amount + 32 bytes detailsHash
+    uint256 private constant TRANSFER_LEN = 129; // 1 byte identifier + 32 bytes recipient + 32 bytes amount + 32 bytes detailsHash + 32 bytes externalId
```
anna-carroll, line 41: this will break things for `Transfer` and `FastTransfer`, need to make a new constant `CONNEXT_TRANSFER_LEN`

anna-carroll: even if deprecating an old message type, you need to make sure that handling for that type still works, or else in-flight messages at the time of upgrade will be borked

packages/contracts-bridge/contracts/BridgeMessage.sol
```diff
@@ -130,22 +131,32 @@ library BridgeMessage {
...
         uint256 _amnt,
-        bytes32 _detailsHash
+        bytes32 _detailsHash,
+        bytes32 _externalId
     ) internal pure returns (bytes29) {
-        return
-            abi.encodePacked(Types.Transfer, _to, _amnt, _detailsHash).ref(0).castTo(
-                uint40(Types.Transfer)
-            );
+        Types transferType = _externalId == bytes32(0) ? Types.Transfer : Types.ConnextTransfer;
```
anna-carroll, line 158: we should not change the content/length/format of `Types.Transfer` - the new Type should not modify the old types - for backwards compatibility

packages/contracts-bridge/contracts/BridgeRouter.sol
```diff
@@ -36,6 +37,8 @@ contract BridgeRouter is Version0, Router {
     ITokenRegistry public tokenRegistry;
     // token transfer prefill ID => LP that pre-filled message to provide fast liquidity
     mapping(bytes32 => address) public liquidityProvider;
+    // reference to connext contract on this domain
+    IConnext public connext;
```
anna-carroll, line 41: if changing storage layout, need to update the Upgrade Gap

packages/contracts-bridge/contracts/BridgeRouter.sol
```diff
@@ -85,10 +88,11 @@ contract BridgeRouter is Version0, Router {
 
     // ======== Initializer ========
 
-    function initialize(address _tokenRegistry, address _xAppConnectionManager)
+    function initialize(address _tokenRegistry, address _xAppConnectionManager, address _connext)
```
anna-carroll, line 91: initialize can only be called once at the proxy's initial deploy. to add new storage variables after the intial deploy, we need a separate setup function. or else, we need to migrate `initialize` to the version that uses numbered initializations instead of binary y/n

anna-carroll (changes requested):
some changes needed for safe upgrade

LayneHaber:
> some changes needed for safe upgrade

Outside of the comments -- do the functional changes look okay at a high level?

packages/contracts-bridge/contracts/BridgeRouter.sol
```diff
@@ -227,6 +207,60 @@ contract BridgeRouter is Version0, Router {
...
+            // query token contract for details and calculate detailsHash
+            _detailsHash = BridgeMessage.getDetailsHash(_t.name(), _t.symbol(), _t.decimals());
+        } else {
+            // if the token originates on a remote chain,
+            // burn the representation tokens on this chain
+            _t.burn(msg.sender, _amount);
+            _detailsHash = _t.detailsHash();
+        }
+        // format Transfer Tokens action
+        bytes29 _action = _externalId == bytes32(0) ? BridgeMessage.formatTransfer(_recipient, _amount, _detailsHash) : BridgeMessage.formatConnextTransfer(_externalId, _amount, _detailsHash);
```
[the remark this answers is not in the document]
anna-carroll: would it be cleaner to have 
```
function _send(
    address _token,
    uint256 _amount,
    uint32 _destination,
    bytes32 _recipientOrExternalId,
    bool _isConnext
) internal {
```
then
```
bytes29 _action = BridgeMessage.formatTransfer(_recipientOrExternalId, _amount, _detailsHash, _isConnext);
```
then
```
function formatConnextTransfer(
    bytes32 _recipientOrExternalId,
    uint256 _amnt,
    bytes32 _detailsHash,
    bool _isConnext
) internal pure returns (bytes29) {
  Type _type = _isConnext ? Types.ConnextTransfer : Types.Transfer;  
  return abi.encodePacked(_type, _recipientOrExternalId, _amnt, _detailsHash).ref(0).castTo(uint40(_type));
}
```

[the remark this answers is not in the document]
anna-carroll: trying to think through which would be clearer for a developer to read / easier to reason about.

packages/contracts-bridge/contracts/BridgeRouter.sol
```diff
@@ -240,38 +274,38 @@ contract BridgeRouter is Version0, Router {
...
         if (_fastEnabled) {
             // If an LP has prefilled this token transfer,
             // send the tokens to the LP instead of the recipient
-            bytes32 _id = BridgeMessage.getPreFillId(
-                _origin,
-                _nonce,
-                _tokenId,
-                _action
-            );
+            bytes32 _id = BridgeMessage.getPreFillId(_origin, _nonce, _tokenId, _action);
```
anna-carroll, line 300: we can deprecate this path in the next upgrade which will make this code less uggo

packages/contracts-bridge/contracts/BridgeRouter.sol
```diff
@@ -290,15 +324,13 @@ contract BridgeRouter is Version0, Router {
             IBridgeToken(_token).setDetailsHash(_action.detailsHash());
         }
         // dust the recipient if appropriate
-        _dust(_recipient);
+        if (_isConnext) {
```
anna-carroll, line 327: thinking through how to reduce these `if/else` branches in the code and reduce sub-functions to a unified path of functionality. we _could_  

- pull `_dust` and `reconcile` calls up into the higher-level `handle` function (after) calling this function)
- remove references to `externalId` from this fn
- pass `recipient` as a param to this function; in higher-up fn, pass `_action.evmRecipient` or `address(connext)` as `recipient` depending on message type

packages/contracts-bridge/contracts/BridgeMessage.sol
```diff
@@ -295,6 +321,8 @@ library BridgeMessage {
      * @param _transferAction The message
      * @return The recipient address as bytes32
      */
+     // NOTE: If used on a Types.ConnextTransfer then this will return the external id. Should there
```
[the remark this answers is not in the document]
anna-carroll: and as-is, adding a type assertion here would break `_handle` because it's called regardless of message type

anna-carroll (commented):
this code should work as-is :) 

review = musings on how to improve it - mainly interested in reducing logical branches for sake of clearer mental models

i will make a PR with some of the changes I discussed

packages/contracts-bridge/contracts/BridgeRouter.sol
```diff
@@ -114,9 +117,11 @@ contract BridgeRouter is Version0, Router {
         bytes29 _action = _msg.action();
         // handle message based on the intended action
         if (_action.isTransfer()) {
-            _handleTransfer(_origin, _nonce, _tokenId, _action, false);
+            _handleTransfer(_origin, _nonce, _tokenId, _action, false, false);
```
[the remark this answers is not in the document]
LayneHaber: definitely is -- main point was easiest way to see the changes based only on existing code paths. not sure the best way to refactor things

packages/contracts-bridge/contracts/BridgeMessage.sol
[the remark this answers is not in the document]
LayneHaber: ah interesting, it seems it would be easy to mess up without it and im in general always in favor of failing loudly as soon as possible

packages/contracts-bridge/contracts/BridgeRouter.sol
```diff
@@ -290,15 +324,13 @@ contract BridgeRouter is Version0, Router {
             IBridgeToken(_token).setDetailsHash(_action.detailsHash());
         }
         // dust the recipient if appropriate
-        _dust(_recipient);
+        if (_isConnext) {
+            connext.reconcile(_externalId, _amount, _canonicalId, _canonicalDomain, _token);
```
anna-carroll, line 328: note to self: does this call ever revert? in what cases?

anna-carroll:
merging this into a branch in the nomad monorepo so I can make edits :)

PR #372, opened 2022-07-05
feat: add forge-std submodule for testing

mattsse:
## Motivation
the `forge-std` submodule was missing causing solc import errors:

```console
Compiler run failed
error[6275]: packages/contracts-core/contracts/test/NomadBase.t.sol:7:1: ParserError: Source "forge-std/console.sol" not found: File outside of allowed directories.
import {console} from "forge-std/console.sol";
```

## Solution
add `forge-std` as submodule via `forge install foundry-rs/forge-std`

**Note*** I had to unignore the `lib` folder (default dir for submodule imports), not sure if this causes side-effects with yarn/javascript builds

if so I can move this to another dir

`FOUNDRY_PROFILE=core forge build ` now works

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 node_modules/
-lib/
 yarn-error.log
 **/tsconfig.tsbuildinfo
 dist/
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "lib/forge-std"]
+	path = lib/forge-std
+	url = https://github.com/foundry-rs/forge-std


mattsse:
> So, we had a looong discussion internally, in an effort not to use git submodules to manage dependencies. The problem we wanted to avoid was that random updates to the library would dirty the git tree.

that's reasonable

gotcha, closing this

prestwich:
we'll probably open a similar PR with the dependency vendored :)

PR #387, opened 2022-07-13
Testing: unit-tests/contracts-bridge

odyslam:
## PR Checklist

- [x] Added Tests
- [ ] Updated Documentation


packages/contracts-bridge/contracts/test/harness/BridgeMessageharness.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity 0.7.6;
```
prestwich, line 2: filename capitalize `Harness`

packages/contracts-bridge/contracts/test/utils/MockHome.sol
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity 0.7.6;
+
+import {Message} from "@nomad-xyz/contracts-core/contracts/libs/Message.sol";
+
+contract MockHome {
```
prestwich, line 6: could this be reused from the `contracts-core` ?
if not, could it be in `contracts-router` so it could be useful to other people implementing routers?

packages/contracts-bridge/contracts/test/BridgeMessage.t.sol
```diff
@@ -0,0 +1,627 @@
...
+        bytes29 tokenId = abi.encodePacked(localDomain, tokenAddress).ref(
+            uint40(BridgeMessage.Types.Invalid)
+        );
+        vm.expectRevert(
+            "Type assertion failed. Got 0x0000000000. Expected 0x0000000001"
+        );
+        BridgeMessage.formatMessage(tokenId, action);
+    }
+
+    function test_formatMessageTransfer() public {
```
prestwich, line 158: some of this is duplicating the logic. instead we should consider hardcoding a set of inputs, and the correct answer. then checking that the logic produces the correct answer

i.e. to test `add(x, y)`, we can hardcode 1,2 and 3, then invoke `add(1,2)` and check that it returns `3`

this is more of a comment on the purpose of unit tests, for use in designing future tests. no specific change needed right now.

odyslam: So, this is something that occurred to me. How I reasoned about it was that I used irrelevant code paths and followed the simple rule of experimentation. Keep everything stable, change 1 parameter. So how I thought about it was like this: "Assuming all code is correct, except the one I am testing". If some of the logic that I use is incorrect, that should come up in it's unit test?

Not sure if that is the correct way to think about it, just how I approached it.

prestwich (approved):

PR #390, opened 2022-07-18
refactor: status/events

ErinHales:
- bug: reference error, fetch undefined (cannot get hosted configs)
 - feature: use indexer to get events

## Motivation

Ability to get status easily and accurately is important for partners trying to integrate with Nomad. We've moved away from this code internally and have not maintained it. We should allow use of indexer for specific queries using the sdk.

## Solution

 1. delete old events code and refactor initializer functions
 2. make an internal `_events()` function that updates a tx's events to the latest and stores them in the cache
 3. make other getter functions for specific events (relay, update, process), call `_events()` only when needed

TODO:
 - [ ] update tests
 - [x] update sdk-bridge
 - [x] check for other downstream changes (indexer, local-env, etc)

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


packages/sdk-bridge/src/bridgeEvents.ts
```diff
@@ -1,29 +0,0 @@
-import { TypedEvent } from '@nomad-xyz/contracts-core';
```
Imti, line 1: 🥳 Yay deleted code

packages/sdk/package.json
```diff
@@ -35,7 +35,10 @@
     "@nomad-xyz/contracts-core": "workspace:^",
     "@nomad-xyz/multi-provider": "workspace:^",
     "axios": "^0.27.2",
```
Imti, line 37: I'd prefer just keeping cross-fetch over axios if that one works better for us. Not a big deal for now tho. Noticed we only use axios in `getProof()`

ErinHales:
Just remembered I need to update the changelog

Imti (approved):
Big refactor, awesome 🌳

packages/sdk/package.json
ErinHales: I just decided to include those changes in this PR, so I'm not mixing code with a release PR

PR #393, opened 2022-07-20
fix: weird promise handling error with checkHomes

ErinHales:
Promise rejection handling was messed up by trying to wrap it in an async forEach block

## Motivation

Was crashing Connext app if they had a bad rpc. They weren't able to handle it on their end.

## Solution

Use for loop instead of forEach

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


Imti (approved):
🎈

yourbuddyconner (approved):

PR #395, opened 2022-07-21
chore: bump nomad packages that deploy uses

odyslam:


anna-carroll (approved):
lgtm - not sure if this is how we should be setting up packages though 👀

PR #396, opened 2022-07-21
fix: explicit cast to string, handles typescript compiler error

odyslam:


anna-carroll (approved):

packages/deploy/src/Contracts.ts
```diff
@@ -21,7 +21,7 @@ export default abstract class Contracts<T> {
   assertIsComplete(): void {
     for (const key of this.keys) {
       if (!this.data[key]) {
-        throw new Error(`Missing key ${key}`);
+        throw new Error(`Missing key ${String(key)}`);
```
anna-carroll, line 24: hmm you could also do `key.toString()`

odyslam: Addressed the comment @anna-carroll . What's the difference?

prestwich: `String(thing)` is a function that coerces the input to `string` using built-in methods.
`thing.toString()` invokes the `toString()` method on that object/prototype. Which may have custom type-defined logic to convert to string. 

using `String` instead of `toString` is discouraged because it potentially skips the type-defined logic

odyslam: What you said makes sense cause of Rust mental models. Thanks!

PR #398, opened 2022-07-21
Fix indexer tx info failure

kekonen:
## Motivation
Indexer in prod and stage recently started falling frequently. The problem localized in getAdditionalInfo method for transaction which gets TX's timestamp, or if that is not present sources it from block. But what if the block is not defined, meaning the TX is not yet mined/confirmed?

## Solution
We skip the events in this case. But It is ok, because we always collect 50+ events from the tip and reapply the events later. If the TX is not confirmed, we just wait for it...

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


yourbuddyconner (approved):

Imti:
Closing since no longer needed

PR #400, opened 2022-07-22
Make keymaster run networks in parallel (concurrently)

kekonen:
## Motivation
Right now keymaster runs nets one by one sequentially, to make it easier to debug. But, we see that it is quite stable, though if there are many networks it takes lots of time to till one net at the end of the queue will get checked and paid.

## Solution
Now we want all nets to have their own loop.

It spins up one "task" per network and the task spins until it breaks in a try-catch loop until an unhandled promise rejection happens, which is impossible to catch anyways. 

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/keymaster/src/keymaster.ts
+++ b/packages/keymaster/src/keymaster.ts
@@ -1,6 +1,7 @@
 import { AgentRole, allowAgent, KeymasterConfig } from "./config";
 import { Accountable, Network, RemoteAgent, RemoteWatcher } from "./account";
 import { Context } from "./context";
+import { sleep } from "./utils";
 
 export class Keymaster {
   config: KeymasterConfig;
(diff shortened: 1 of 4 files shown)


packages/keymaster/src/keymaster.ts
```diff
@@ -90,20 +91,27 @@ export class Keymaster {
...
     networks: string[],
+    period: number,
     dryrun = false
   ): Promise<void> {
-    for (const net of this.networks.values()) {
-      try {
-        if (networks.length === 0 || networks.includes(net.name)) {
-          await net.checkAndPay(dryrun);
+    const promises = Array.from(this.networks.values()).map(async (net) => {
+      while (true) {
```
Imti, line 98: Is this `while(true)` needed here?

Imti: nvm I see now

yourbuddyconner: Please add a comment here explaining what this `.map` with a `while(true)` is being used for, as it's relatively unintuitive at first glance.

kekonen: Added comment

yourbuddyconner (approved):

PR #404, opened 2022-07-26
Feat: Use yarn cache to improve CI speed

odyslam:


prestwich (changes requested):
add `--frozen-lockfile` to `yarn install` and see if it improves the install step further

PR #414, opened 2022-07-29
feat: add token sender to TransferToHook message

anna-carroll:
## Motivation
Routers that build on top of Nomad Home need to authenticate the sender of the message is a valid Router on the other chain. Similarly, Routers that build on top of Nomad BridgeRouter need to do the same.

## Solution
Add the sender of the tokens + message on the origin chain to the TransferToHook message format.

## PR Checklist

- [x] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-bridge/contracts/BridgeMessage.sol
+++ b/packages/contracts-bridge/contracts/BridgeMessage.sol
@@ -411,7 +411,7 @@ library BridgeMessage {
     /**
      * @notice Retrieves the extra data from a TransferWithHook
      * @param _transferAction The message
-     * @return The extra data as a bytes29
+     * @return A TypedMemview of extraData
      */
     function extraData(bytes29 _transferAction)
         internal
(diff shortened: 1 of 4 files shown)


packages/contracts-bridge/contracts/BridgeMessage.sol
```diff
@@ -388,6 +393,26 @@ library BridgeMessage {
...
+        returns (bytes32)
+    {
+        // before = 1 byte identifier + 32 bytes hook address + 32 bytes amount + 32 bytes detailsHash = 97
+        return _transferAction.index(97, 32);
+    }
+
+    /**
+     * @notice Retrieves the extra data from a TransferWithHook
+     * @param _transferAction The message
+     * @return The extra data as a bytes29
```
prestwich, line 414: nit: a typed pointer to the extra data in memory

prestwich (approved):
approved with nit

packages/contracts-bridge/contracts/BridgeMessage.sol
anna-carroll: https://github.com/nomad-xyz/monorepo/pull/414/commits/75b3d2d83f8c8fce1c06471a2fcdee2c488f9dee

PR #431, opened 2022-08-22
unit-test: Message.t.sol

odyslam:


packages/contracts-core/contracts/test/Message.t.sol
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    function test_body() public {
+        messageView = message.ref(0);
+        assertEq(Message.body(messageView).keccak(), body.ref(0).keccak());
+    }
+
+    function test_leaf() public {
+        messageView = message.ref(0);
+        assertEq(Message.leaf(messageView), messageHash);
```
prestwich, line 109: method syntax is preferred

```
messageView.leaf()
```

odyslam:
actually, let me do all

packages/contracts-core/contracts/test/Message.t.sol
```diff
@@ -0,0 +1,114 @@
...
+                recipient,
+                body
+            ),
+            messageHash
+        );
+    }
+
+    function test_origin() public {
+        messageView = message.ref(0);
+        assertEq(uint256(messageView.origin()), uint256(originDomain));
```
anna-carroll, line 79: why casting them to `uint256`?  is this necessary for `assertEq` function?

anna-carroll (approved):
LGTM

packages/contracts-core/contracts/test/Message.t.sol
odyslam: yeah, at some point i want to PR some signatures to support other types of uints

prestwich (approved):

PR #437, opened 2022-08-31
unit-tests(bridgeToken.sol)

odyslam:
Based on Solmate's ERC20 [tests](https://github.com/transmissions11/solmate/blob/main/src/test/ERC20.t.sol), with some refactoring around cheats and bridgeToken specifics.


packages/contracts-bridge/package.json
```diff
@@ -40,7 +40,8 @@
     "@nomad-xyz/contracts-router": "workspace:^",
     "@openzeppelin/contracts": "~3.4.2",
     "@openzeppelin/contracts-upgradeable": "~3.4.2",
-    "@summa-tx/memview-sol": "^2.0.0"
+    "@summa-tx/memview-sol": "^2.0.0",
+    "solmate": "^6.5.0"
```
prestwich, line 44: is this used?

packages/contracts-bridge/contracts/test/BridgeToken.t.sol
```diff
@@ -0,0 +1,687 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity 0.7.6;
+
+import {MockWeth} from "./utils/MockWeth.sol";
+import "forge-std/Test.sol";
+
+// Tests are largely based on ERC20Test from solmate
```
prestwich, line 7: we can't include code directly from solmate due to license restrictions. Can you link the specific files that you referenced so we can check the overlap?

odyslam: Most of the tests that test the ERC20 functions that are inherited and used by BridgeToken.

prestwich: Can you link the specific files that you referenced so we can check the overlap?

prestwich: is comment now redundant?

prestwich (approved):

PR #440, opened 2022-09-01
Avail bridge router implementation

gretzke:
This pull request includes the implementation for the nomad bridge router and corresponding message declaration with Polygon Avail


packages/contracts-da-bridge/contracts/test/DABridgeMessage.t.sol
```diff
@@ -0,0 +1,127 @@
...
+/*//////////////////////////////////////////////////////////////
+                            LIBRARIES
+//////////////////////////////////////////////////////////////*/
+
+import "forge-std/Test.sol";
+import "forge-std/console2.sol";
+import {TypedMemView} from "@summa-tx/memview-sol/contracts/TypedMemView.sol";
+import {TypeCasts} from "@nomad-xyz/contracts-core/contracts/libs/TypeCasts.sol";
+
+contract DABridgeMessageTest is Test {
```
luketchang, line 19: @odyslam could you take a pass on the tests for the message and router files?

luketchang (commented):
Minus some minor things, overall looks lgtm but want @anna-carroll @prestwich @odyslam to take a pass as well :)

packages/contracts-da-bridge/contracts/DABridgeRouter.sol
```diff
@@ -0,0 +1,128 @@
...
+        uint32 _nonce,
+        bytes32 _sender,
+        bytes memory _message
+    ) external override onlyReplica onlyRemoteRouter(_origin, _sender) {
+        bytes29 _view = _message.ref(0).getTypedView();
+        if (_view.isValidDataRoot()) {
+            _handleDataRoot(_origin, _nonce, _view);
+        } else {
+            revert("!valid message");
+        }
```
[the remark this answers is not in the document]
anna-carroll: yeah this should be fine

packages/contracts-da-bridge/contracts/DABridgeRouter.sol
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * @notice emitted when a new data root is received
+     * @param originAndNonce Domain where the transfer originated and the
+     *        unique identifier for the message from origin to destination,
+     *        combined in a single field ((origin << 32) & nonce)
+     * @param blockNumber for data root
+     * @param root data root
+     */
+    event Receive(
+        uint64 indexed originAndNonce,
```
anna-carroll, line 42: what is the purpose of passing origin and nonce here? why not just origin? is nonce used by any off-chain applications in this case?

packages/contracts-da-bridge/contracts/DABridgeRouter.sol
```diff
@@ -0,0 +1,128 @@
...
+     * @param _message The message in the form of raw bytes
+     */
+    function _handleDataRoot(
+        uint32 _origin,
+        uint32 _nonce,
+        bytes29 _message
+    ) internal {
+        (uint64 blockNumber, bytes32 dataRoot) = _parse(_message);
+        assert(roots[blockNumber] == 0);
+        roots[blockNumber] = dataRoot;
```
anna-carroll, line 93: is this app meant to be receiving roots from only one chain - Avail?
if not, what if two of the same `blockNumber` is passed for different chains?

anna-carroll: if roots are only meant to be coming from one chain, then should there be some guardrails to ensure that remote routers are only enrolled for that one domain? or that messages are only coming from that one domain?

luketchang: > is this app meant to be receiving roots from only one chain - Avail?

Yes this is correct

luketchang: > should there be some guardrails to ensure that remote routers are only enrolled for that one domain?

I assumed we would just handle this in the connection manager. But you're saying we might want an additional check at the app level? I personally don't think we should hardcode this kind of logic into the app and just leave the connection manager to do what it's meant for

anna-carroll: the Router (this contract) is the one that manages registering remote routers. the connection manager registers Replicas.

so, if you wanted to limit this contract from only receiving messages from one single remote router, you would need to only ever register one single remote router. you can't override the function `enrollRemoteRouter` unfortunately, so you'd need to ensure that the deploy script never gets this wrong. that's somewhat brittle, ideally the contract would ensure it can't be used incorrectly to the extent possible.

one example of how you could do this (in a gas efficient way) would be to add an immutable variable in the constructor, `availDomain` (or something more generally named) and require in `handle` that `origin == availDomain`

anna-carroll: another question: what is the thinking behind possible re-orgs? is that not a concern with Avail's consensus mechanism? what if the `root` at `blockNumber` changes, and the app wants to overwrite it

packages/contracts-da-bridge/contracts/DABridgeRouter.sol
anna-carroll: ah i guess they are using it to link the `Receive` event definitively to the message that caused it, since this is the unique identifier for a message. you should be able to just observe the Process(messageHash) event that fires directly after Receive, but fair enough if you want it explicitly linked within this data

packages/contracts-da-bridge/contracts/DABridgeRouter.sol
luketchang: > you'd need to ensure that the deploy script never gets this wrong. that's somewhat brittle

Yes agreed this may make it worth it to encode the remote domain check into the app. @gretzke would you mind pushing this change? As Anna noted, we'd need to add an immutable `uint32` called `availDomain` that we set in the constructor and then add another modifier to the `handle` method that checks `origin == availDomain`

gretzke: Since this is an upgradeable contract I can't set an immutable variable. I pushed changes with a normal variable

anna-carroll: @gretzke what do you mean? you can set immutables on upgradable contracts

gretzke: Immutable variables have to be set in the constructor and upgradeable contracts (usually) don't have constructors. Many upgrade frameworks (e.g., open zeppelin) explicitly prohibit the existence of constructors in upgradeable contracts for that reason. Also the variable wouldn't point to a proxy storage slot but to the variable in the bytecode of the implementation contract. So in case of a contract upgrade the immutable variable would have to be set again in the constructor allowing for error where the variable is set to another value (or the constructor is deleted and the variable remains uninitialized). Therefore in my opinion it's not safe to use immutable variables in upgradeable contracts. But it's up to you, if you wish for the variable to be immutable I can change that.

anna-carroll: hmm okay!! i don't feel toooo strongly about it, though the gas increase is regrettable.

i will say
> the constructor is deleted and the variable remains uninitialized
in this case, the contract wouldn't compile


> in case of a contract upgrade the immutable variable would have to be set again in the constructor allowing for error
true, though if variable ordering is re-arranged that's another vector for error when performing an upgrade. i'll allow, though, that the error you cite would probably be in deploy code, which is often not audited, whereas changing storage slots would be in protocol code, which usually undergoes more scrutiny

anna-carroll: three options would be 
- set `availDomain` as storage var, check on `handle`
- set `availDomain` as immutable var, check on `handle`
- do not check domain at all in `handle`. ensure that `remoteRouter` mapping  is configured s.t. only one domain can ever send to the router 

honestly, all of these are probably fine outcomes.

luketchang (approved):

luketchang:
Closed in favor of #473

PR #443, opened 2022-09-02
fix: MerkleTest hex conversion

odyslam:
## PR Checklist

- [x] Added Tests
- [x] Updated Documentation

### DIFF
--- a/packages/contracts-core/contracts/test/utils/MerkleTest.sol
+++ b/packages/contracts-core/contracts/test/utils/MerkleTest.sol
@@ -18,12 +18,14 @@ contract MerkleTest is Test {
             bytes32[32] memory
         )
     {
+        // Hash the message
+        string memory hash = toHexString(keccak256(message));
         string[] memory input = new string[](7);
         input[0] = "yarn";
         input[1] = "gen-proof";
         input[2] = "-r";
         input[3] = "-m";
-        input[4] = toHexString(uint256(keccak256(message)));
+        input[4] = hash;
         input[5] = "-i";
         input[6] = "0";
         bytes memory result = vm.ffi(input);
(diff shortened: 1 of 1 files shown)


packages/contracts-core/contracts/test/utils/MerkleTest.sol
```diff
@@ -36,34 +38,119 @@ contract MerkleTest is Test {
         return (root, leaf, index, proof);
     }
 
-    // Pulled from String.sol from OpenZeppelin
-    // source: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.6.0/contracts/utils/Strings.sol
-    function toHexString(uint256 value) internal pure returns (string memory) {
-        if (value == 0) {
-            return "0x00";
+    /**
```
prestwich, line 41: btw, the more efficient version is in `packages/contracts-bridge/contracts/Encoding.sol` 😅

packages/contracts-core/contracts/test/utils/MerkleTest.sol
```diff
@@ -36,34 +38,119 @@ contract MerkleTest is Test {
...
-        return toHexString(value, length);
     }
 
-    function toHexString(uint256 value, uint256 length)
+    function toHexString(bytes32 input) private returns (string memory) {
+        (uint256 a, uint256 b) = encodeHex(uint256(input));
+        return string(abi.encodePacked("0x", a, b));
+    }
+
+    function toUint256(bytes memory _bytes)
```
prestwich, line 147: this will have potentially unintended results if the bytearray is less than 32 bytes long

odyslam:
ok, will switch to using the one from Encoding.sol in the contracts-bridge package

packages/contracts-core/contracts/test/utils/MerkleTest.sol
prestwich: this function is still broken. also appears to be unused

odyslam: removed it. my bad. Why would it have unintended results if the byte array is less than 32 bytes long?

prestwich:
why are tests broken? o.O

packages/contracts-core/contracts/test/utils/MerkleTest.sol
odyslam: ah makes sense. Thanks

prestwich (approved):

PR #456, opened 2022-09-08
feat: support the use of arbitrary shell commands in the nomad-cli

Imti:
## Motivation

Killswitch will be written in rust. We should support multiple languages via the use of running arbitrary shell commands that we can use to run rust binaries. This makes the cli more flexible.

## Solution

Tried a couple different ways such as passing shell commands directly to `run` and also adding a bash script as an oclif command. Ended up with using `execa` as a way to manage child processes and piping the ouput of that child process to the parent process for visibility. This has the added benefit of supporting error handling.

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


rswanson (approved):
lgtm

PR #458, opened 2022-09-13
feat: LE unit testing improvements; mainnet forking

xBalbinus:
## Motivation

Lots of things:
- 'Unit tests' (not really) used to rely on Docker and used to pull containers from the cloud. Now they no longer do. We have achieved > 50% coverage on all unit-testable items.
- We want to be able to fork mainnet and be able to write tests accordingly (for end to end tests on contract upgrades, contract to contract calls, etc.). Local environment is now capable of supporting integration tests on a mainnet fork. *CAVEAT: There will be additional refactoring of LE for being able to test by reading from config files of existing Nomad deployments. Current integration tests are not designed to be done on a forked mainnet. Additional tests that fit mainnet not included within this PR and will most likely be included in a future one.
- Needed to refactor the whole local environment to accept Jest's Docker mocks.

## Solution

- Refactored all unit tests of local environment - unit tests no longer pull Docker containers from the cloud.
- Implemented (and tested) local environment's capability to now interact with **live contract instances** (see wETH used as instantiation)
- Mainnet forked, hardhat funded accounts. Of course, mainnet doesn't have to be just ETH (any Alchemy-supported network)
- Jest Docker mocks implemented on container creation (testing whether or not all modules are able to interact with Docker)
- Docker as optional input into localenv instantiation
- CI Time for all integration tests: (~1200 seconds). LE Unit tests reduced to sub 30 seconds.
- Cleaned up abstractions (included a new abstraction, ForkedNetwork to represent mainnet forks).

## PR Checklist

- [x] Added Tests
- [x] Updated Documentation (incoming in a separate document)


xBalbinus:
```
unknown account #0 (operation="getAddress", code=UNSUPPORTED_OPERATION, version=providers/5.6.8)

      at Logger.Object.<anonymous>.Logger.makeError (../../node_modules/@ethersproject/logger/src.ts/index.ts:261:28)
      at Logger.Object.<anonymous>.Logger.throwError (../../node_modules/@ethersproject/logger/src.ts/index.ts:273:20)
      at ../../node_modules/@ethersproject/providers/src.ts/json-rpc-provider.ts:204:24
```

Does anyone understand why I'm getting this? From some initial searches it seems that the deploy script's ethers.getAddress() is not recognized as a supported function on mainnet forks? Any guidance would be significantly appreciated.

packages/local-environment/src/actor.ts
```diff
@@ -24,17 +24,17 @@ export abstract class Actor {
 
 export abstract class DockerizedActor extends Actor {
   
-  docker: Docker;
-  container?: Docker.Container;
+  docker: Dockerode;
```
kekonen, line 27: Docker not used inside. No need it to be here

packages/local-environment/hardhat/hardhat.config.js
```diff
@@ -56,5 +64,9 @@ module.exports = {
       // url: "http://0.0.0.0:8545",
       accounts,
     },
+    forking: {
+      url: "https://eth-mainnet.g.alchemy.com/v2/${process.env.ALCHEMY_API_KEY}",
+      blockNumber: parseInt(process.env.BLOCK_NUMBER),
+    },
```
kekonen, line 70: Create a temp object, add forking into it on demand, pass it to the module.exports

packages/local-environment/src/domain.ts
```diff
@@ -24,15 +25,29 @@ export class NomadDomain {
...
   metricsPort?: number;
   nomadEnv?: NomadEnv;
+  forkUrl?: string;
+  forkBlockheight?: number;
+  docker: Dockerrode;
 
   connectedNetworks: NomadDomain[];
 
-  constructor(network: Network, nomadEnv?: NomadEnv) {
+  constructor(name: string, domain: number, forkUrl?: string, docker?: Dockerrode, nomadEnv?: NomadEnv) {
```
kekonen, line 34: I don't like that Docker is moved inside of NomadDomain. I think there should be another level of abstraction. In particular, forkUrl and block height is only used to create another network within the constructor and then it is stored into the domain for some reason.

If we need docker (and other stuff) in here u need to extend the Domain class with those traits. This will allow u keeping the original Domain being constructed from a network.

```
class DockerForkedDomain extends NomadDomain {
  constructor(name, domain, ....)
    const network = new HardhatOrForkedNetwork_ForInstance(...);
    super(network,...)
```

packages/local-environment/src/network.ts
```diff
@@ -32,31 +32,44 @@ export abstract class Network {
...
+    ...args: string[]
+  ): Promise<ethers.Contract>;
 
   constructor(
     name: string,
     domainNumber: number,
     chainId: number,
-    blockTime = 10000
+    docker: Dockerode,
+    forkUrl?: string
```
kekonen, line 72: Same as domain comment. `Network` is lower level stuff. Not the place for docker

packages/local-environment/src/network.ts
```diff
@@ -340,15 +362,203 @@ export class HardhatNetwork extends Network {
...
...60003373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff1681526020019081526020016000205410151515610a2757600080fd5b80600360003373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff168152602001908152602001600020600082825403925050819055503373ffffffffffffffffffffffffffffffffffffffff166108fc829081150290604051600060405180830381858888f193505050501515610ab457600080fd5b3373ffffffffffffffffffffffffffffffffffffffff167f7fcf532c15f0a6db0bd6d0e038bea71d30d808c7d98cb3bf7268a95bf5081b65826040518082815260200191505060405180910390a250565b600260009054906101000a900460ff1681565b60036020528060005260406000206000915090505481565b60018054600181600116156101000203166002900480601f016020809104026020016040519081016040528092919081815260200182805460018160011615610100020316600290048015610bc65780601f10610b9b57610100808354040283529160200191610bc6565b820191906000526020600020905b815481529060010190602001808311610ba957829003601f168201915b505050505081565b6000610bdb33848461068c565b905092915050565b60046020528160005260406000206020528060005260406000206000915091505054815600a165627a7a72305820deb4c2ccab3c2fdca32ab3f46728389c2fe2c165d5fafa07661e4e004f6c344a0029";
```
kekonen, line 484: Please move out these long hardcodings into a separate file. I cannot even select them multiple lines for the review :)

kekonen (dismissed):
I see that you try to incorporate docker into multiple core abstractions. This is not correct. First, make sure you actually need them there, second, please add another level of abstraction, so that we don't incorporate docker specific problems into stuff that has nothing to do with docker at all ;)

kekonen:
Also in general, docker in many places is a required constructor parameter. Dockerode in general is a very much a thing that could, in theory, be initialized as many times as we want as it does connect to global docker. This means that mostly the parameter can be optional (even if it is required in the struct), so docker can be instantiated if the param is not present. This can be helpful if we don't construct the whole env, but use components on their own, like for unit test or something similar

packages/local-environment/src/actor.ts
```diff
@@ -24,17 +24,17 @@ export abstract class Actor {
...
   logMatcher?: StreamMatcher;
 
   log = bunyan.createLogger({ name: "localenv" });
 
-  constructor(name: string, actorType: string) {
-    super(name, actorType)
-    this.docker = new Docker();
+  constructor(name: string, actorType: string, docker: Dockerode) {
+    super(name, actorType);
+    this.docker = docker;
```
kekonen, line 37: Probably a optional parameter. if not present fallback to `new Dockerode()`

packages/local-environment/hardhat/hardhat.config.js
xBalbinus: Hardhat has now been made into a temp object, and forking added on demand.

packages/local-environment/src/actor.ts
xBalbinus: Addressed

packages/local-environment/src/domain.ts
xBalbinus: As prescribed, DockerForkedDomain will now be a new abstraction completely SEPARATE from the default hardhat networks. Hardhat networks will now be purely for localhosted testing - alternative settings for docker, hardhat forking etc will go in this DockerForkedDomain abstraction.

xBalbinus: Note: DockerForkedDomain will extend NomadDomain.

kekonen:
All my comments above are stupid. Only thing is required is to adjust HardhatNetwork for these optional params.

yourbuddyconner (approved):

packages/local-environment/.env.example
```diff
@@ -5,8 +5,18 @@ PRIVATE_KEY_3=3000000000000000000000000000000000000000000000000000000000000003
...
 
-AGENTS_IMAGE=gcr.io/nomad-xyz/nomad-agent:prestwich-remove-deploy-gas
\ No newline at end of file
+AGENTS_IMAGE=gcr.io/nomad-xyz/nomad-agent:prestwich-remove-deploy-gas
+
+WETH_ADDRESS=
+
+# Only set the following if you're forking mainnet
+
+ALCHEMY_API_KEY=UEz-5kiUuyGqZhi8SNUrgJrEy-LPqkZK
```
Imti, line 18: @xBalbinus Just heads up that we should remove this api key and rotate it too just in case

xBalbinus: Got it np

packages/local-environment/src/domain.ts
```diff
@@ -41,27 +54,28 @@ export class NomadDomain {
...
-  connectNetwork(d: NomadDomain) {
+  connectNetwork(d: NomadDomain): void {
     if (!this.connections().includes(d.network.name))
       this.connectedNetworks.push(d);
   }
 
-  async isAgentsUp(): Promise<boolean> {
-    return await this.agents!.isAllUp();
+  async isAgentsUp(): Promise<boolean | undefined> {
+    return await this.agents?.isAllUp();
```
Imti, line 78: If `this.agents` is not defined, does that mean that they're still spinning up?

packages/local-environment/src/network.ts
```diff
@@ -175,19 +189,20 @@ export class HardhatNetwork extends Network {
   keys: Key[];
   handler: DockerizedNetworkActor;
 
-  constructor(name: string, domain: number, options?: HardhatNetworkOptions) {
-    super(name, domain, options?.chainId || domain);
-    this.handler = options?.handler || new DockerizedNetworkActor(this.name);
+  constructor(name: string, domain: number, docker?: Dockerode, options?: HardhatNetworkOptions) {
+    super(name, domain, domain, docker);
```
Imti, line 193: I noticed we're passing in `domain` twice, is that right? Before it looks like we passed in `options?.chainId || domain` for the third parameter of super

packages/local-environment/src/network.ts
```diff
@@ -340,15 +359,164 @@ export class HardhatNetwork extends Network {
...
+  blockTime: number;
+  keys: Key[];
+  blockExplorer: string;
+  firstStart: boolean;
+  handler: DockerizedNetworkActor;
+  weth: string;
+  forkUrl: string;
+
+  constructor(name: string, domain: number, forkUrl: string, wETHAddress: string, docker?: Dockerode, blockExplorer?: string, blocktime?: number, options?: HardhatNetworkOptions) {
+    super(name, domain, domain, docker);
```
Imti, line 384: Same here, passing in domain twice

packages/local-environment/src/network.ts
```diff
@@ -175,19 +189,20 @@ export class HardhatNetwork extends Network {
...
-  constructor(name: string, domain: number, options?: HardhatNetworkOptions) {
-    super(name, domain, options?.chainId || domain);
-    this.handler = options?.handler || new DockerizedNetworkActor(this.name);
+  constructor(name: string, domain: number, docker?: Dockerode, options?: HardhatNetworkOptions) {
+    super(name, domain, domain, docker);
+    this.handler = options?.handler || new DockerizedNetworkActor(this.name, this.docker);
     this.blockTime = 10;
     this.firstStart = false;
     this.keys = options?.keys || [];
+    this.rpc = [`http://localhost:${this.handler.port}`];
```
Imti, line 198: Has to be localhost?

packages/local-environment/src/network.ts
xBalbinus: Fixed this just now

packages/local-environment/src/network.ts
```diff
@@ -340,15 +359,164 @@ export class HardhatNetwork extends Network {
...
+    return provider.getSigner(addressOrIndex);
+  }
+
+  getJsonRpcProvider(): ethers.providers.JsonRpcProvider {
+    return new ethers.providers.JsonRpcProvider(
+      this.rpc[0]
+    );
+  }
+
+  async isConnected(): Promise<boolean> {
```
Imti, line 485: Does this function need to be async? I noticed we're not awaiting `this.handle.isConnected()` below

packages/local-environment/src/domain.ts
xBalbinus: No this just checks docker status for all agents.

packages/local-environment/src/network.ts
xBalbinus: We're about to remove this

packages/local-environment/src/network.ts
xBalbinus: Removing this

packages/local-environment/src/network.ts
xBalbinus: Yes, all mainnet forking happens on localhost - it's just a hardhat config.

packages/local-environment/src/domain.ts
Imti: Just realized we might wanna use `areAgentsUp` as opposed to `isAgentsUp` lol

packages/local-environment/tests/actor-agent.unit.test.ts
```diff
@@ -0,0 +1,85 @@
+import { NomadDomain } from "../src/domain";
+import { LocalAgent, AgentType } from "../src/agent";
+import Dockerode from 'dockerode';
+
+// jest.mock("dockerode");
```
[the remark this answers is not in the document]
xBalbinus: Removed

packages/local-environment/tests/nomadenv.unit.test.ts
```diff
@@ -1,95 +1,51 @@
...
+        id: "0x" + "20".repeat(20),
+    });
+    
+    expect(le.refreshSDK(le.nomadConfig)).toBeDefined();
+});
+
+test("Can create deployContext", async () => {
+
+    const tDomain = new NomadDomain('tom', 1);
+    const le = new NomadEnv({
```
Imti, line 45: Quick suggestion, we can use `beforeEach` to instantiate `le` before each test.

packages/local-environment/src/utils.ts
```diff
@@ -1,11 +1,12 @@
...
 import { ethers } from "ethers";
 
-export function sleep(ms: number) {
+export function sleep(ms: number): Promise<void> {
   return new Promise((res) => setTimeout(res, ms));
 }
 
 export function readLine(q: string): Promise<string> {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
   const readline = require("readline");
```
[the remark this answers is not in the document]
kekonen: I don't remember the implication, but there was some reason I think. But not 100% sure. Probably lets fix later, outside of the scope of this PR

kekonen (changes requested):

packages/local-environment/src/domain.ts
```diff
@@ -41,27 +49,33 @@ export class NomadDomain {
       watcher: signer,
       recoveryManager: signer,
     });
+
+  }
+
+  static newHardhatNetwork(name: string, domain: number, creationParams?: DomainOptions): NomadDomain {
```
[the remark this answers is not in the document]
xBalbinus: I low key like the current naming - at least fits into context when I'm writing docs... lmk if you come up with something better than `newWithHardhatNetwork` though :P no offense

packages/local-environment/src/domain.ts
```diff
@@ -28,11 +36,11 @@ export class NomadDomain {
   connectedNetworks: NomadDomain[];
 
   constructor(network: Network, nomadEnv?: NomadEnv) {
+    this.network = network;
     this.connectedNetworks = [];
     this.keys = new AgentKeys();
-    this.network = network;
 
-    this.nomadEnv = nomadEnv;
+    if (nomadEnv) this.nomadEnv = nomadEnv;
```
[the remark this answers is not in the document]
xBalbinus: Removed the if clause

packages/local-environment/.env.example
kekonen, line 18: this still needs to get removed

packages/local-environment/.env.example
```diff
@@ -5,8 +5,18 @@ PRIVATE_KEY_3=3000000000000000000000000000000000000000000000000000000000000003
...
+BLOCK_NUMBER=15670813
+
+tDomainNumber=1
+jDomainNumber=2
```
kekonen, line 22: Why do we need to have those in `.env`? or generally set up outside of the test case scenario?

packages/local-environment/src/network.ts
```diff
@@ -39,24 +39,33 @@ export abstract class Network {
   recoveryManager: string;
   weth: string;
 
+  rpc: string[];
```
kekonen, line 42: lets make rpc a getter method `get rpcs(): string[]`. I see it as a dynamic param. Fix the following code according to this. In particular no need to have a variable holding it from construction.

kekonen: wait, we already have that. then u can remove the rpc field from network, and make sure that Implementations of Network just query `.rpcs`

packages/local-environment/src/nomadenv.ts
```diff
@@ -25,16 +25,19 @@ export class NomadEnv {
...
-  refreshSDK(config: NomadConfig) {
+  refreshSDK(config: NomadConfig): BridgeContext {
     this.bridgeSDK = new BridgeContext(config);
+    this.coreSDK = new NomadContext(config);
+    return this.bridgeSDK;
   }
 
   // Adds a network to the array of networks if it's not already there.
-  addDomain(d: NomadDomain) {
+  addDomain(network: Network): void {
```
kekonen, line 39: Why would `addDomain` method require a network Xiangan???
AddDomain adds a domain. AddNetwork would add network.

packages/local-environment/src/nomadenv.ts
```diff
@@ -212,76 +238,91 @@ export class NomadEnv {
...
-  // Await domains to up networks.
-  await Promise.all([tDomain.network.up(), jDomain.network.up()]);
+  if (process.env.tDomainNumber) {
+    tDomainNumber = parseInt(process.env.tDomainNumber);
+  }
 
-  log.info(`Upped Tom and Jerry`);
+  if (process.env.jDomainNumber) {
+    jDomainNumber = parseInt(process.env.jDomainNumber);
+  }
```
kekonen, line 280: This. Why do we need to fetch domain from somewhere else? In this case we would maybe also want to fetch domain's name from somewhere else too? (the last question is not a call to action, but rather a call to thinking, but first question might be a call to discussion)

packages/local-environment/src/domain.ts
```diff
@@ -41,27 +49,33 @@ export class NomadDomain {
...
-    if (!this.agents) this.agents = new Agents(this, metricsPort, this.nomadEnv);
+  ensureAgents(metricsPort = 9090, docker?: Dockerode): void {
+    if (!this.agents) this.agents = new Agents(this, metricsPort, docker, this.nomadEnv);
   }
 
-  connectNetwork(d: NomadDomain) {
+  connectNetwork(d: NomadDomain): void {
     if (!this.connections().includes(d.network.name))
       this.connectedNetworks.push(d);
   }
```
kekonen, line 75: Maybe add a parameter `connectInReverse`:
```
  connectNetwork(d: NomadDomain, connectInReturn=true): void {
    if (!this.connections().includes(d.network.name)) {
      this.connectedNetworks.push(d);
      if (connectInReturn) {
         d.connectInReturn(this, false)
      }
    }
  }
```
Question to think: how will it help us?

kekonen (changes requested):
Some more thingies

packages/local-environment/.env.example
xBalbinus: In case of network forking, users will then be able to put in a domain number that matches with the real chain ID (but it's honestly not that important and can be left out - lmk if you think this is the better option)

packages/local-environment/src/nomadenv.ts
xBalbinus: ^ See above, not sure if you think removing this is optimal (in consideration for network forking to replicate the real chain ID

packages/local-environment/src/network.ts
xBalbinus: Removed.

packages/local-environment/.env.example
xBalbinus: Removed

packages/local-environment/src/nomadenv.ts
xBalbinus: Renaming :)

packages/local-environment/src/domain.ts
xBalbinus: Reduces the redundancy in code that forces us to do connectNetwork twice for each domain. Implementing.

xBalbinus:
Rebased

packages/local-environment/.env.example
kekonen: lets get rid of it please. It just create additional fuzz. If they want they can set it in particular test, especially that they will define forking in that test.

packages/local-environment/src/nomadenv.ts
kekonen: it is not a chainId, it is domain number, which is only according to Nomad/or user. It has nothing to do with the chainId

packages/local-environment/src/domain.ts
```diff
@@ -41,27 +49,45 @@ export class NomadDomain {
...
 
-  async isAgentsUp(): Promise<boolean> {
-    return await this.agents!.isAllUp();
+  connectInReturn(d: NomadDomain): boolean {
+    if(this.connections().includes(d.network.name)) {
+      return false;
+    }
+    else return true;
+  }
```
kekonen, line 86: no, need for this. Please, could u try to do it as it was in my example with additional param (that has default value)

packages/local-environment/src/nomadenv.ts
```diff
@@ -141,6 +145,23 @@ export class NomadEnv {
     return false;
   }
 
+  get forkUrl(): string | undefined {
+    if (process.env.ALCHEMY_API_KEY)
+      return `https://eth-mainnet.g.alchemy.com/v2/${process.env.ALCHEMY_API_KEY}`;
+  }
```
kekonen, line 151: What if Alchemy key is provided, but we work with not a mainnet? Also I don't think it has anything to do with a NomadEnv in general. It is purely a Network layer thing. In particular HardhatNetwork

packages/local-environment/src/nomadenv.ts
```diff
@@ -141,6 +145,23 @@ export class NomadEnv {
...
+
+  get wETHAddress(): string {
+    return `${process.env.WETH_ADDRESS}`;
+  }
```
kekonen, line 155: Also this, weth not guaranteed to have the same address in all nets.

packages/local-environment/src/nomadenv.ts
```diff
@@ -141,6 +145,23 @@ export class NomadEnv {
...
+
+  get tDomain(): NomadDomain | undefined {
+    return this.domains[0];
+  }
+
+  get jDomain(): NomadDomain | undefined {
+    return this.domains[1];
+  }
+
```
kekonen, line 164: Also these guys. T and J should only exist in the Test case, they should not migrate into the business logic

kekonen (changes requested):

packages/local-environment/src/nomadenv.ts
xBalbinus: Ill replace with the fork url in .env

packages/local-environment/.env.example
xBalbinus: Removed

packages/local-environment/src/nomadenv.ts
xBalbinus: Removed

packages/local-environment/src/nomadenv.ts
kekonen: But also it is probably not the place for it in the first place. What if we want to have Tom a forked net and Jerry hardhat from scratch or something similar.

packages/local-environment/src/nomadenv.ts
xBalbinus: Removed

packages/local-environment/src/nomadenv.ts
xBalbinus: Removed / replaced with weth

xBalbinus: *from process.env on creation

packages/local-environment/src/nomadenv.ts
xBalbinus: Okay then, removing

packages/local-environment/src/domain.ts
xBalbinus: Revised

packages/local-environment/src/nomadenv.ts
xBalbinus: Removed

packages/local-environment/src/nomadenv.ts
xBalbinus: Removed

packages/local-environment/tests/nomadenv.int.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        
+        let tDomainNumber = 1;
+        let jDomainNumber = 2;
+      
+        if (process.env.tDomainNumber) {
+          tDomainNumber = parseInt(process.env.tDomainNumber);
+        }
+      
+        if (process.env.jDomainNumber) {
+          jDomainNumber = parseInt(process.env.jDomainNumber);
```
kekonen, line 28: I thought we deleted jDomainNumber from .env

kekonen (changes requested):

kekonen (approved):
Just fix my additional comment, and go once more through your changes. Otherwise lgtm

packages/local-environment/tests/nomadenv.int.test.ts
xBalbinus: Oh shoot sorry about that - yes it should be deleted already, removing this.

xBalbinus: Tysm!

xBalbinus:
Got it, will fix this ASAP - thanks for the feedback.

In light of this, I will separate the removal of yarn boot up into a
different PR.

James8T:
https://github.com/nomad-xyz/monorepo/issues/483
Investigation on Nomad CTO and bounty campaign on Nomad founders and its VCs and partners backers #483

James8T:
https://github.com/nomad-xyz/monorepo/issues/483

PR #460, opened 2022-09-16
feat: UpdaterManager tests

odyslam:


packages/contracts-core/contracts/test/UpdaterManager.t.sol
```diff
@@ -0,0 +1,90 @@
...
+        emit FakeSlashed(reporter);
+        vm.prank(address(home));
+        updaterManager.slashUpdater(reporter);
+    }
+
+    function test_renounceOwnership() public {
+        updaterManager.renounceOwnership();
+        // Gas metered how much gas is the no-op
+        // if we change it from no-op, the gas consumption will change
+        assertEq(gasleft(), 9223372036854747223);
```
prestwich, line 88: wouldn't it be more clear to test that the owner doesn't change?

prestwich (approved):

PR #466, opened 2022-09-19
bugfix: fixing release workflow and version for local-environment 

rswanson:
## Motivation

Fixing the versioning for local-environment and fixing up the publish script. Should publish now

## Solution

Realized that the `npm publish` command needed to be local to the local-environment package directory, so I fixed that in the workflow. 

## PR Checklist

- [x] Added Tests
- [x] Updated Documentation


Imti (approved):
:octocat:

PR #467, opened 2022-09-19
bugfix: final local-environment release changes

rswanson:
## Motivation

Testing github actions is a bit of a slog. Found one last bug plus a few optimizations on the commands we're running. 

## Solution

After updating the flow I realized that running the `yarn version` command in the pipeline can lead to some unexpected behavior later, so I'm going to remove it in lieu of releasing the version thats in the branch at the time. 

## PR Checklist

- [x] Added Tests
- [x] Updated Documentation


Imti (approved):
🦘

PR #469, opened 2022-09-19
test: Unit and Fuzzing tests for GovernanceRouter

anna-carroll:
## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


packages/contracts-core/contracts/test/utils/MockXAppConnectionManager.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
prestwich, line 1: move to contracts-router, as it'll be useful to anything downstream of there

prestwich: oh i see, you're using it in local tests, nvm 🤔

PR #478, opened 2022-10-08
chore: remove gas snapshot from CI

odyslam:
- We don't gas gold and it was producing more noise than signal


prestwich (approved):

James8T:
ETHmad holders will have recovery near 0%. Only 1-2 hackers hold the stolen ETH, probably will never been returned.

I want you JAMES PRESTWICH and all Nomad team to contribute with own pocket, you have $ETH!!!. I cant pay scholarships to my son because of you!!! Also Why don’t you negociate with big 8 figures wallets to write-off their debt against equity?

PR #495, opened 2022-10-31
Prestwich/sol config

prestwich:
Allows scripts to load and parse the Nomad Config from solidity

## Motivation

Speed development of upgrade scripts & related work

## Solution

provide `Config.sol` which loads a config from a file, and exposes solidity accessors

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-core/CHANGELOG.md
+++ b/packages/contracts-core/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Unreleased
 
+- feature: Config.sol exposes configuration to foundry scripts
 - (patch): filter scripts from hh compilation
 
 ### 2.0.0
--- a/packages/contracts-core/CHANGELOG.md
+++ b/packages/contracts-core/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ### Unreleased
 
-- feature: Config.sol exposes configuration to foundry scripts
 - (patch): filter scripts from hh compilation
 
 ### 2.0.0
--- a/packages/contracts-core/contracts/scripts/Config.sol
+++ b/packages/contracts-core/contracts/scripts/Config.sol
@@ -9,12 +9,9 @@ import {UpdaterManager} from "../UpdaterManager.sol";
 import {XAppConnectionManager} from "../XAppConnectionManager.sol";
 
 import "forge-std/Vm.sol";
-import "forge-std/StdJson.sol";
 import "forge-std/Test.sol";
 
 abstract contract Config {
-    using stdJson for string;
-
     Vm private constant vm =
         Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
 
--- a/packages/contracts-core/contracts/scripts/Config.sol
+++ b/packages/contracts-core/contracts/scripts/Config.sol
@@ -1,0 +1,214 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity 0.7.6;
+pragma abicoder v2;
+
+import {UpgradeBeaconProxy} from "../upgrade/UpgradeBeaconProxy.sol";
+import {UpgradeBeacon} from "../upgrade/UpgradeBeacon.sol";
+import {UpgradeBeaconController} from "../upgrade/UpgradeBeaconController.sol";
+import {UpdaterManager} from "../UpdaterManager.sol";
+import {XAppConnectionManager} from "../XAppConnectionManager.sol";
+
+import "forge-std/Vm.sol";
+import "forge-std/StdJson.sol";
+import "forge-std/Test.sol";
+
+abstract contract Config {
+    using stdJson for string;
+
+    Vm private constant vm =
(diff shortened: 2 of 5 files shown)


packages/contracts-core/contracts/scripts/Config.sol
```diff
@@ -0,0 +1,214 @@
...
+    }
+}
+
+contract TestJson is Test, Config {
+    function setUp() public {
+        config = '{"core": {"ethereum": {"deployHeight": 1234, "governanceRouter": {"proxy":"0x569D80f7FC17316B4C83f072b92EF37B72819DE0","implementation":"0x569D80f7FC17316B4C83f072b92EF37B72819DE0","beacon":"0x569D80f7FC17316B4C83f072b92EF37B72819DE0"}}}}';
+    }
+
+    function test_Json() public {
+        require(coreDeployHeight("ethereum") == 1234);
```
odyslam, line 208: Why use require instead of assert?

odyslam: I would suggest to use assert to keep the code idiomatic

packages/contracts-core/contracts/scripts/Config.sol
```diff
@@ -0,0 +1,214 @@
...
+import {UpgradeBeaconController} from "../upgrade/UpgradeBeaconController.sol";
+import {UpdaterManager} from "../UpdaterManager.sol";
+import {XAppConnectionManager} from "../XAppConnectionManager.sol";
+
+import "forge-std/Vm.sol";
+import "forge-std/StdJson.sol";
+import "forge-std/Test.sol";
+
+abstract contract Config {
+    using stdJson for string;
```
odyslam, line 16: I think that you don't use the library at all

packages/contracts-core/contracts/scripts/Config.sol
```diff
@@ -0,0 +1,214 @@
...
+    function __Config_initialize(string memory file) internal {
+        require(!isInitialized(), "already init");
+        config = vm.readFile(file);
+    }
+
+    function isInitialized() public view returns (bool) {
+        return bytes(config).length != 0;
+    }
+
+    function core(string memory domain) private returns (string memory) {
```
odyslam, line 43: prob cleaner to do `config.readString(path)` and use the stdJson helpers?

If you agree, we should do it everywhere.

prestwich: I don't think that does the same thing. that reads a string from the json value, while this returns the entire object.

that said, this method is redundant and unused, and ought to be removed

packages/contracts-ops/contracts/Config.sol
```diff
@@ -0,0 +1,202 @@
...
+import {UpgradeBeaconProxy} from "@nomad-xyz/contracts-core/contracts/upgrade/UpgradeBeaconProxy.sol";
+import {UpgradeBeacon} from "@nomad-xyz/contracts-core/contracts/upgrade/UpgradeBeacon.sol";
+import {UpgradeBeaconController} from "@nomad-xyz/contracts-core/contracts/upgrade/UpgradeBeaconController.sol";
+import {UpdaterManager} from "@nomad-xyz/contracts-core/contracts/UpdaterManager.sol";
+import {XAppConnectionManager} from "@nomad-xyz/contracts-core/contracts/XAppConnectionManager.sol";
+
+import "forge-std/Vm.sol";
+import "forge-std/Test.sol";
+
+abstract contract Config {
```
anna-carroll, line 14: qq: is the intention to use this in conjunction with the `NomadProtocol` contract structure? Asking because the intention behind that was to be able to load fork tests (via config) AND locally deployed integration tests within the same interface, so tests could be easily used for both (RFC: https://app.clickup.com/14205595/v/dc/dhgmv-26427/dhgmv-25167 )

anna-carroll: if i understand correctly, the entire purpose of this class is parsing values from the config in a clean way - it'll be used in the fork fixture to instantiate `NomadProtocol` from a config?

anna-carroll: (perhaps can add a little documentation to the top of the contract as well)

packages/contracts-core/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### Unreleased
 
+- feature: Config.sol exposes configuration to foundry scripts
```
anna-carroll, line 5: this is `contracts-core` changelog

anna-carroll (commented):
fabbbb i love it

packages/contracts-ops/contracts/Config.sol
prestwich: gonna document all in the readme, with updater rotation as an example

anna-carroll (approved):

PR #497, opened 2022-11-01
Prestwich/updater rotation

prestwich:
## Motivation

Enable updater rotation via recovery manager

## Solution

A Simple forge script that outputs encoded calls that can be input to a callbatch

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


packages/contracts-ops/contracts/RotateUpdater.sol
```diff
@@ -0,0 +1,59 @@
...
+        string[] memory connections = connections(localDomain);
+        address newUpdater = updater(localDomain);
+
+        Home home = home(localDomain);
+        UpdaterManager updaterManager = updaterManager(localDomain);
+
+        if (newUpdater != home.updater()) {
+            push(
+                address(home),
+                abi.encodeWithSelector(home.setUpdater.selector, newUpdater)
```
prestwich, line 32: is this correct? my reading of Home is that this will fail, but I documented this as necessary. cc @anna-carroll

packages/contracts-core/contracts/scripts/CallBatch.sol
```diff
@@ -0,0 +1,117 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
anna-carroll, line 1: i think this file shouldn't be in two places?

packages/contracts-ops/README.md
```diff
@@ -2,6 +2,31 @@
 
 Contains forge scripts for Nomad system maintenance
 
+## Building Blocks
+
+- `Config.sol` utilities for loading and parsing Nomad Config JSON
+  - initalize the config with the path to a JSON config
+- `CallBatch.sol` utility for aggregating calls into a callbatch, and
```
anna-carroll, line 9: could be useful to include a bit more contextual information about what a callbatch is. even "into a governance callbatch" could be helpful

packages/contracts-ops/README.md
```diff
@@ -2,6 +2,31 @@
...
+  outputting the batch to JSON
+  - initialize a callbatch with the local domain name and an output path
+  - output paths will be prepended with `./actions/`
+  - call `push(to, data)` to add a call to the batch
+  - call `finish()` to output the list of calls
+  - call `build()` to output the built call to `executeGovernanceActions()`
```
anna-carroll, line 15: note: we'd like two build options - normal governance & recovery

packages/contracts-ops/README.md
```diff
@@ -2,6 +2,31 @@
...
+
+## Scripts & Entrypoints
+
+- `RotateUpdaters.sol`
+  - Rotates updaters of all
```
anna-carroll, line 20: incomplete sentence

packages/contracts-ops/contracts/CallBatch.sol
```diff
@@ -15,7 +17,9 @@ abstract contract CallBatch is Script {
     string public domain;
     string public outputFile;
 
-    function initializeCallBatch(
+    GovernanceMessage.Call built;
```
anna-carroll, line 20: in a built callbatch, i believe `to` should be an `address` not a `bytes32`

packages/contracts-ops/contracts/CallBatch.sol
```diff
@@ -98,14 +101,34 @@ abstract contract CallBatch is Script {
...
+        require(!complete, "already written");
         complete = true;
         writeOutput();
     }
+
+    function build(address governanceRouter) public {
+        require(bytes(domain).length != 0, "must initialize");
+        require(bytes(outputFile).length != 0, "must initialize");
+        require(!complete, "already written");
+        complete = true;
```
anna-carroll, line 113: hmmm so you can only call `finish` or `build` - not both?
also - should finish be renamed something like `output` or `write`?

packages/contracts-ops/contracts/CallBatch.sol
```diff
@@ -98,14 +101,34 @@ abstract contract CallBatch is Script {
     function finish() public {
         require(bytes(domain).length != 0, "must initialize");
         require(bytes(outputFile).length != 0, "must initialize");
+        require(!complete, "already written");
         complete = true;
         writeOutput();
     }
+
+    function build(address governanceRouter) public {
```
anna-carroll, line 109: this always builds for recovery mode. won't work for normal chainops actions. this relates, ofc, to the comment about presently only having calls for one domain in this library. i am open to punting & refactoring this later when we need it, though=

packages/contracts-ops/contracts/CallBatch.sol
anna-carroll: why don't we just combine the functionalities of `finish` and `build` - writing the un-built and the built calldata to a file each time, similar to the typescript?

packages/contracts-ops/contracts/Config.sol
```diff
@@ -122,6 +128,24 @@ abstract contract Config {
             );
     }
 
+    function replicaOfUpdgrade(string memory local, string memory replica)
```
anna-carroll, line 131: typo in function name

anna-carroll: wdyt rename `replica` to `remote` to follow convention with sdk more closely?

packages/contracts-ops/contracts/Config.sol
```diff
@@ -80,14 +82,18 @@ abstract contract Config {
...
-    function home(string memory domain)
+    function homeUpgrade(string memory domain)
         public
         onlyInitialized
         returns (Upgrade memory)
     {
         return abi.decode(loadCoreAttribute(domain, "home"), (Upgrade));
     }
 
+    function home(string memory domain) public onlyInitialized returns (Home) {
```
anna-carroll, line 93: if you're adding these for Home and Replica, should also add for GovernanceRouter, BridgeRouter, and TokenRegistry?

packages/contracts-ops/contracts/scripts/RotateUpdater.s.sol
```diff
@@ -0,0 +1,87 @@
...
+        string calldata localDomain,
+        string calldata output
+    ) public {
+        initialize(configFile, localDomain, output);
+        setUpdater();
+        finish();
+    }
+
+    // entrypoint
+    function createRecoveryTx(
```
anna-carroll, line 78: yeah so wdyt about just having one entrypoint and combining the functionality of `finish` and `build`

packages/contracts-ops/contracts/scripts/RotateUpdater.s.sol
```diff
@@ -0,0 +1,87 @@
...
+import {UpdaterManager} from "@nomad-xyz/contracts-core/contracts/UpdaterManager.sol";
+import {Home} from "@nomad-xyz/contracts-core/contracts/Home.sol";
+import {Replica} from "@nomad-xyz/contracts-core/contracts/Replica.sol";
+
+import "forge-std/Script.sol";
+
+contract RotateUpdater is Script, Config, CallBatch {
+    function setReplicaUpdater(
+        string memory localDomain,
+        string memory connection
```
anna-carroll, line 17: rename `remoteDomain` for consistency?

packages/contracts-ops/README.md
prestwich: I think the appropriate place for that is the governance documentation, not the toolbox

packages/contracts-ops/README.md
prestwich: yes, that's the behavior described here

packages/contracts-ops/contracts/CallBatch.sol
prestwich: because you'll never want both. either you want to combine the list with other lists and THEN build for governance (the `finish()` call), or you want to `build()` now for recovery

packages/contracts-ops/contracts/CallBatch.sol
prestwich: that is the intentional design

packages/contracts-ops/contracts/Config.sol
prestwich: meh

packages/contracts-ops/contracts/Config.sol
prestwich: yes, as they become needed

packages/contracts-ops/contracts/scripts/RotateUpdater.s.sol
prestwich: see above, there's no situation where we want to do that, and devs should know what they want to do

https://github.com/nomad-xyz/monorepo/pull/497#discussion_r1010520205

packages/contracts-ops/contracts/RotateUpdater.sol
anna-carroll: calling directly would fail. need to call through UpdaterManager, as you've changed the code to do :)

packages/contracts-ops/README.md
anna-carroll: 🤔  no this only builds for recovery. normal governance puts all domains together

packages/contracts-ops/contracts/CallBatch.sol
anna-carroll: mm when we used the typescript library, i often used the ability to look at the broken down calls in addition to the built calldata. to sanity check that the correct number of calls were there, etc.

packages/contracts-ops/contracts/Config.sol
```diff
@@ -122,6 +128,24 @@ abstract contract Config {
             );
     }
 
+    function replicaOfUpdgrade(string memory local, string memory remote)
```
anna-carroll, line 131: still typo

packages/contracts-ops/contracts/CallBatch.sol
prestwich: we do that because TS is junk and we want to know we wrote it write. much lower chance of error doing it in solidity

packages/contracts-ops/contracts/Config.sol
anna-carroll: https://github.com/nomad-xyz/monorepo/pull/497/commits/07cce7153c0c2787ddc0ca9d79fafc3f428db20c

packages/contracts-ops/contracts/CallBatch.sol
anna-carroll: @prestwich the `built` is copy-pasted into Gnosis multisig transactions. if the `to` is a `bytes32` then it's going to be a problem. we have multiple addresses which start with zeros, too, so cutting leading zeros isn't a good solution.

prestwich: built is not used, so it's been removed. the output writes an address (see the `vm.toString` call in `build()`

odyslam (approved):

PR #499, opened 2022-11-01
feat: mock accountant for testing recovery UI

prestwich:
## Motivation

Enable UI testing of recovery flows

## Solution

A mock accountant that implements most basic flows

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-bridge/contracts/mocks/MockAccountant.sol
+++ b/packages/contracts-bridge/contracts/mocks/MockAccountant.sol
@@ -8,6 +8,15 @@ contract MockNftAccountant is NFTRecoveryAccountant {
 
     constructor() NFTRecoveryAccountant(address(0), address(1)) {}
 
+    function exposed_increaseTotalAffected(address _asset) public {
+        if (totalAffected[_asset] == 0) {
+            totalAffected[_asset] = 5000 ether;
+        } else {
+            totalAffected[_asset] = totalAffected[_asset] + 10 ether;
+        }
+
+    }
+
     function exposed_record(
         address _asset,
         address _user,
--- a/packages/contracts-bridge/contracts/mocks/MockAccountant.sol
+++ b/packages/contracts-bridge/contracts/mocks/MockAccountant.sol
@@ -35,6 +35,7 @@ contract MockNftAccountant is NFTRecoveryAccountant {
         require(_user == msg.sender, "only NFT holder can recover");
         Record memory _rec = records[_id];
         uint256 _amount = 50000000;
+        _rec.recovered += _amount;
         emit Recovery(_id, _rec.asset, _user, _amount);
     }
 
--- a/packages/contracts-bridge/contracts/mocks/MockAccountant.sol
+++ b/packages/contracts-bridge/contracts/mocks/MockAccountant.sol
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity 0.7.6;
+
+import {NFTRecoveryAccountant} from "../accountants/NFTAccountant.sol";
+
+contract MockNftAccountant is NFTRecoveryAccountant {
+    mapping(address => bool) public allowList;
+
+    constructor() NFTRecoveryAccountant(address(0), address(1)) {}
+
+    function exposed_record(
+        address _asset,
+        address _user,
+        uint256 _amount
+    ) external {
+        _record(_asset, _user, _amount);
+    }
+
+    modifier onlyAllowed() {
+        require(allowList[msg.sender], "not allowed");
+        _;
+    }
+
+    function recover(uint256 _id) external onlyAllowed {
+        address _user = ownerOf(_id);
+        require(_user == msg.sender, "only NFT holder can recover");
+        Record memory _rec = records[_id];
+        uint256 _amount = 50000000;
+        emit Recovery(_id, _rec.asset, _user, _amount);
+    }
+
+    function allow(address who) external {
+        allowList[who] = true;
+    }
+
+    function disallow(address who) external {
+        allowList[who] = false;
+    }
+
+    function allow(address[] calldata _who) external {
+        for (uint256 i = 0; i < _who.length; i++) {
+            allowList[_who[i]] = true;
+        }
+    }
+
+    function disallow(address[] calldata _who) external {
+        for (uint256 i = 0; i < _who.length; i++) {
+            allowList[_who[i]] = false;
+        }
+    }
+
+    function info(uint256 _id)
+        external
+        view
+        returns (
+            address _holder,
+            bool _isAllowed,
+            string memory _uri,
+            address _asset,
+            uint256 _originalAmount,
+            address _originalUser,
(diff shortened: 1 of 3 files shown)


packages/contracts-bridge/contracts/mocks/MockAccountant.sol
```diff
@@ -0,0 +1,76 @@
...
+pragma solidity 0.7.6;
+
+import {NFTRecoveryAccountant} from "../accountants/NFTAccountant.sol";
+
+contract MockNftAccountant is NFTRecoveryAccountant {
+    mapping(address => bool) public allowList;
+
+    constructor() NFTRecoveryAccountant(address(0), address(1)) {}
+
+    function exposed_record(
```
anna-carroll, line 11: this will fail (`"overmint"`) unless we set `totalAffected` for some testnet tokens. 

it would be nice to set `totalAffected` anyway, as that's used in the GUI logic as well i think

anna-carroll (changes requested):
need to set `totalAffected`

packages/contracts-bridge/contracts/mocks/MockAccountant.sol
prestwich: gonna add a public function to increase total affected

packages/contracts-bridge/contracts/mocks/MockAccountant.sol
```diff
@@ -0,0 +1,76 @@
...
+    modifier onlyAllowed() {
+        require(allowList[msg.sender], "not allowed");
+        _;
+    }
+
+    function recover(uint256 _id) external onlyAllowed {
+        address _user = ownerOf(_id);
+        require(_user == msg.sender, "only NFT holder can recover");
+        Record memory _rec = records[_id];
+        uint256 _amount = 50000000;
```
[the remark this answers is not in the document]
anna-carroll: (for the gui)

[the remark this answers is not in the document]
prestwich: we're not really testing that functionality

packages/contracts-bridge/contracts/mocks/MockAccountant.sol
```diff
@@ -0,0 +1,76 @@
...
+    {
+        _holder = ownerOf(_id);
+        _isAllowed = allowList[_holder];
+        _uri = tokenURI(_id);
+        Record memory _rec = records[_id];
+        _asset = _rec.asset;
+        _originalAmount = uint256(_rec.amount);
+        _originalUser = _rec.originalUser;
+        _recovered = uint256(_rec.recovered);
+        _recoverable = 50000000;
```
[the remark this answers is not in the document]
prestwich: we're also not testing this functionality

packages/contracts-bridge/contracts/mocks/MockAccountant.sol
ErinHales: I don't think I need totalAffected

packages/contracts-bridge/contracts/mocks/MockAccountant.sol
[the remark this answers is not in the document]
ErinHales: Yeah, would just use `info` because we'll want to get the asset and amount at least as well

packages/contracts-bridge/contracts/mocks/MockAccountant.sol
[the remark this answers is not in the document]
ErinHales: Incrementing `_rec.recovered` would be great

ErinHales:
Will bridging both directions be enabled in staging? Or only bridging back, like production?

ErinHales:
It sounds like we're planning to use Goerli and Sepolia in staging (alchemy has stopped supporting kovan, ropsten, rinkeby, etc).  We won't have any existing representations between these chains, so we would need the ability to bridge and create new representations at least initially

packages/contracts-bridge/contracts/mocks/MockAccountant.sol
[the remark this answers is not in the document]
prestwich: done

prestwich:
> Will bridging both directions be enabled in staging? Or only bridging back, like production?

this contract will not be connected to a bridge. NFTs can be minted using `exposed_record`

packages/contracts-bridge/contracts/mocks/MockAccountant.sol
```diff
@@ -0,0 +1,85 @@
...
+        _;
+    }
+
+    function recover(uint256 _id) external onlyAllowed {
+        address _user = ownerOf(_id);
+        require(_user == msg.sender, "only NFT holder can recover");
+        Record memory _rec = records[_id];
+        uint256 _amount = 50000000;
+        _rec.recovered += uint96(_amount);
+        emit Recovery(_id, _rec.asset, _user, _amount);
```
ErinHales, line 38: Ah, so this doesn't actually send any funds, it just records it and emits an event

ErinHales (approved):

packages/contracts-bridge/contracts/mocks/MockAccountant.sol
```diff
@@ -0,0 +1,85 @@
...
+    modifier onlyAllowed() {
+        require(allowList[msg.sender], "not allowed");
+        _;
+    }
+
+    function recover(uint256 _id) external onlyAllowed {
+        address _user = ownerOf(_id);
+        require(_user == msg.sender, "only NFT holder can recover");
+        Record memory _rec = records[_id];
+        uint256 _amount = 50000000;
```
anna-carroll, line 36: nit: this is in wei, right? so 0.00000000005 ETH - since this is for the purpose of developing the frontend, preferable to use a more readable value like `0.05 ether`

anna-carroll:
Documenting - 

for setup: 
- get some tokenAddress to test with - can be a production token address, a testnet token, or a nonsense address 
- call `exposed_increaseTotalAffected(tokenAddress)`
- call `exposed_record(tokenAddress, testUser, someAmount)`
- once you want testUser to be on the allowList (if you've already tested state when they're not on the list) call `allow(testUser)`

packages/contracts-bridge/contracts/mocks/MockAccountant.sol
```diff
@@ -0,0 +1,85 @@
...
+    {
+        _holder = ownerOf(_id);
+        _isAllowed = allowList[_holder];
+        _uri = tokenURI(_id);
+        Record memory _rec = records[_id];
+        _asset = _rec.asset;
+        _originalAmount = uint256(_rec.amount);
+        _originalUser = _rec.originalUser;
+        _recovered = uint256(_rec.recovered);
+        _recoverable = 50000000;
```
anna-carroll, line 83: should also change this to `0.05 ether` now too, yes?

prestwich: doesn't matter

PR #503, opened 2022-11-02
feat: upgrade pipeline

odyslam:
## Solution
- Forge script that deploys new implementations on multiple chains and updates the config file
- Forge script that reads the updated config file and produces the Upgrade callBatches, one for every chain. It can also produce the built combined callBatch in case we execute every upgrade on every chain independently (e.g recovery mode)
- The `actions` directory top-level directory is used for input/output files
- Requires: https://github.com/foundry-rs/foundry/pull/3647 and https://github.com/foundry-rs/forge-std/pull/213

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


foundry.toml
```diff
@@ -1,16 +1,16 @@
 [profile.default]
-solc_version = '0.7.6'        # Version of solc that we use
```
prestwich, line 2: why all the formatting changes?

packages/contracts-ops/contracts/DeployImplementations.s.sol
```diff
@@ -0,0 +1,221 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
prestwich, line 1: file exists twice? once in `contracts/` and  once in `contracts/scripts/`

prestwich:
as a procedural question, is there any way we can avoid deploying the implementation if it's a NOP? i.e. the implementation code would not change? it seems difficult as we use immutables 🤔

packages/contracts-ops/contracts/DeployImplementations.s.sol
```diff
@@ -0,0 +1,221 @@
...
+
+    function title(string memory title1, string memory title2) internal {
+        console2.log(" ");
+        console2.log("===========================");
+        console2.log(title1, title2);
+        console2.log("===========================");
+        console2.log(" ");
+    }
+
+    function writeImpl() internal {
```
prestwich, line 156: refactor to use JsonWriter?

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
```diff
@@ -0,0 +1,221 @@
...
+        // initializeImplementations();
+        writeImpl();
+        vm.stopBroadcast();
+    }
+
+    /*//////////////////////////////////////////////////////////////
+                      IMPLEMENTATION DEPLOYMENT
+  //////////////////////////////////////////////////////////////*/
+
+    function deployImplementations() internal {
```
prestwich, line 91: clean

packages/contracts-ops/contracts/scripts/RotateUpdater.s.sol
```diff
@@ -82,6 +82,6 @@ contract RotateUpdater is Script, Config, CallBatch {
     ) public {
         initialize(configFile, localDomain, output);
         setUpdater();
-        build(address(governanceRouter(localDomain).proxy));
+        build(address(governanceRouterUpgrade(localDomain).proxy));
```
prestwich, line 85: change should be rebased out

packages/contracts-ops/scripts/updateConfig.ts
```diff
@@ -0,0 +1,77 @@
+import * as config from "@nomad-xyz/configuration";
```
prestwich, line 1: I like this but think it needs more development. what are the common actions, and do we have standard formats for them?

I like having the implementations interface 🤔

packages/contracts-ops/contracts/scripts/UpgradeGoVBatches.s.sol
```diff
@@ -0,0 +1,219 @@
...
+// Ops libs
+import {Config} from "../Config.sol";
+import {CallBatch} from "../Callbatch.sol";
+
+contract UpgradeGovBatches is Test, Config, CallBatch {
+    /*//////////////////////////////////////////////////////////////
+                                 BEACONS
+    //////////////////////////////////////////////////////////////*/
+
+    address homeBeacon;
```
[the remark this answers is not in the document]
odyslam: What do you mean?

packages/contracts-ops/contracts/DeployImplementations.s.sol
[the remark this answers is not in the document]
odyslam: This is for console output, not file/json

foundry.toml
odyslam: vim

packages/contracts-ops/contracts/DeployImplementations.s.sol
odyslam: yy not sure why. I don't see this locally.

packages/contracts-ops/contracts/DeployImplementations.s.sol
odyslam: Yes

foundry.toml
prestwich: yeah, but can we not

packages/contracts-ops/contracts/CallBatch.sol
```diff
@@ -44,92 +44,128 @@ abstract contract CallBatch is Script {
...
+
+    function writeObjectBody(string memory indent, string[2][] memory kvs)
+        private
+    {
+        for (uint256 i = 0; i < kvs.length; i++) {
+            writeKV(indent, kvs[i][0], kvs[i][1], i == kvs.length - 1);
+        }
+    }
+
+    function writeArrayOpen(string memory indent, string memory key) private {
```
anna-carroll, line 78: why is this in CallBatch? needs to be rebased out, this should be in JSONWriter, no?

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
```diff
@@ -0,0 +1,244 @@
...
+            loadConfig();
+            vm.startBroadcast();
+            deployImplementations();
+            updateImpl(localDomain);
+            vm.stopBroadcast();
+        }
+    }
+
+    function loadConfig() internal {
+        recoveryTimelock = getGovernanceConfiguration(localDomain)
```
anna-carroll, line 83: just use calls to config inline, it's less boilerplate

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
```diff
@@ -0,0 +1,244 @@
...
+import {JsonWriter} from "../JsonWriter.sol";
+
+contract DeployImplementations is Test, Config {
+    using JsonWriter for JsonWriter.Buffer;
+
+    /*//////////////////////////////////////////////////////////////
+                            UPGRADE ARGUMENTS
+  //////////////////////////////////////////////////////////////*/
+
+    // configuration
```
anna-carroll, line 31: just use this from Config - no?

odyslam:
We should first merge #516 , rebase this and then merge.  Review only the deployImplementations script

packages/contracts-ops/contracts/CallBatch.sol
odyslam: Yeah, ignore this file. It will be rebased.

odyslam:
I had to break the fork creation and fork selection into two different phases because of a forge bug (that was just fixed in https://github.com/foundry-rs/foundry/pull/3679)

The script works as it should.

Next step is to open a PR with the upgradeCallbatches and update the readme on how all this should be used.

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
```diff
@@ -0,0 +1,254 @@
...
+        // GovernanceRouter
+        governanceRouter = new GovernanceRouter(
+            localDomainNumber,
+            recoveryTimelock
+        );
+        governanceRouter.initialize(xAppConnectionManager, address(0));
+        console2.log("Governance Router deploed at", address(governanceRouter));
+
+        // BridgeRouter
+        bridgeRouter = new BridgeRouter();
```
anna-carroll, line 152: what about `EthereumBridgeRouter` for ethereum?

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
```diff
@@ -0,0 +1,254 @@
...
+        vm.writeJson(vm.toString(address(home)), configPath, valueKey);
+        valueKey = string(
+            abi.encodePacked(
+                ".core.",
+                domain,
+                ".governanceRouter",
+                ".implementation"
+            )
+        );
+        vm.writeJson(
```
anna-carroll, line 180: this does not compile

odyslam: It just needs the latest forge-std

odyslam: Which I made a different PR for

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
anna-carroll: @odyslam if we wanted to merge this at some point, would have to take this into account. but by now, i think we should move forward with my modifications of your work, as it's closer to mergeable. you can refactor / edit that work (And the other scripts) to make them multi-chain and use `writeJson` - but we shouldn't block timelines on that.

PR #507, opened 2022-11-03
feature: prankable callbatch

prestwich:
## Motivation

Allow forktests to prank scripts

## Solution

Add prank() and prankRecovery() to the CallBatch
These will execute the callbatch against governance, when running in anvil

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-ops/contracts/CallBatch.sol
+++ b/packages/contracts-ops/contracts/CallBatch.sol
@@ -100,6 +100,27 @@ abstract contract CallBatch is Script {
         buffer.writeSimpleObject("", "", kvs, true);
         buffer.flushTo(outputFile);
     }
+
+    function prank(address router) public {
+        // prank the router itself to avoid governor chain issues
+        vm.prank(router);
+        GovernanceRouter(router).executeGovernanceActions(
+            calls,
+            new uint32[](0),
+            new GovernanceMessage.Call[][](0)
+        );
+    }
+
+    function prankRecovery(address router) public {
+        // prank recovery (only works if recovery is active)
+        address recovery = GovernanceRouter(router).recoveryManager();
+        vm.prank(recovery);
+        GovernanceRouter(router).executeGovernanceActions(
+            calls,
+            new uint32[](0),
+            new GovernanceMessage.Call[][](0)
+        );
+    }
 }
 
 contract TestCallBatch is CallBatch {
--- a/packages/contracts-ops/contracts/CallBatch.sol
+++ b/packages/contracts-ops/contracts/CallBatch.sol
@@ -101,7 +101,7 @@ abstract contract CallBatch is Script {
         buffer.flushTo(outputFile);
     }
 
-    function prank(address router) public {
+    function prankExecuteBatch(address router) public {
         // prank the router itself to avoid governor chain issues
         vm.prank(router);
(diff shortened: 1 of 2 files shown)


packages/contracts-ops/contracts/CallBatch.sol
```diff
@@ -161,6 +160,27 @@ abstract contract CallBatch is Script {
         kvs[1][1] = vm.toString(data);
         writeSimpleObject("", "", kvs, true);
     }
+
+    function prank(address router) public {
```
anna-carroll, line 164: i would maybe consider `prankExecuteBatch()` or something like that since `CallBatch` will be in the inheritance tree, it's a little clearer what's happening as opposed to `vm.prank`

packages/contracts-ops/contracts/CallBatch.sol
```diff
@@ -161,6 +160,27 @@ abstract contract CallBatch is Script {
...
+        );
+    }
+
+    function prankRecovery(address router) public {
+        // prank recovery (only works if recovery is active)
+        address recovery = GovernanceRouter(router).recoveryManager();
+        vm.prank(recovery);
+        GovernanceRouter(router).executeGovernanceActions(
+            calls,
+            new uint32[](0),
```
anna-carroll, line 180: something initially confusing for me about this library is that it's *only* a CallBatch for recovery mode. might be helpful to include that in the documentation (inline, or README) and/or naming. initially confusing because CallBatch in the SDK is a different purpose

anna-carroll (approved):
lgtm - just some thoughts on clarity for dev users (naming, documentation)

packages/contracts-ops/contracts/CallBatch.sol
prestwich: agreed 👍

packages/contracts-ops/contracts/CallBatch.sol
prestwich: well, it's ONLY a callbatch for 1 chain. it can be used for non-recovery mode

anna-carroll: yeah, but the only chain that can be used as 1 chain (Excepting recovery mode) is the Governor router. and when the Governor router is not in recovery mode, it's meant to be able to execute remote calls. so effectively it's a recovery mode CallBatch

packages/contracts-ops/contracts/CallBatch.sol
anna-carroll: https://github.com/nomad-xyz/monorepo/pull/507/commits/e7e8a5ad769985a70d6629c7e6561ccaa470f864

packages/contracts-ops/contracts/CallBatch.sol
prestwich: the current conception is that JSON outputs could be combined out of band

prestwich: i.e. when running for governance, you 


```
$ forge run Script --sig run(string) moonbeam.json
$ forge run Script --sig run(string) ethereum.json
$ jq -s '.[0] * .[1]' ./actions/ethereum.json ./actions/moonbeam.json
```

PR #512, opened 2022-11-07
totalAffected Calculations

xBalbinus:
### Methodology

- for each asset, `totalAffected` is equal to the total outstanding bridged amount. 
- this equals 
    - (1) the amount of that asset ever bridged out to other chains, 
    - minus (2) the total amount of that asset ever bridged in to its origin domain
- the resulting figure is the total outstanding amount it is still possible to bridge in
- total amount bridged out can be determined by 
    - querying all Send events where token = asset address and taking the sum of amount field
- total amount bridged in can be determined by 
    - querying all Receive events where token = asset address and taking the sum of amount field

## Solution

Attached in the PR's changelog is a GRAPHQL file demonstrating the queries currently used to get to the target numbers.
This is to be deleted upon merge.

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


temp.graphql
```diff
@@ -0,0 +1,1046 @@
+query Mainnet {
+  #162633775975 - 162633775975 = 0
+  tok0sent: subgraph_send_aggregate(where: {token: {_eq: "0x2260fac5e5542a773aa44fbcfedf7c193bc2c599"}, _gs_chain: {_eq: "mainnet"}}, distinct_on: id) {
```
anna-carroll, line 3: it should be possible to `groupBy: token` (in the same section where you're currently doing "aggregate") so you can have a single query for `send` and a single query for `receive`. then you can probably even take it a step further and filter something like `where: token: {_in: ["0xtoken1..., 0xtoken2..."]}`

anna-carroll: if there are only two queries, it's easier to review each of them very closely. but since each of the tokens' queries are identical, it's also easy to put on blinders and just review one token's query as the general case :)

anna-carroll (commented):
some thoughts on the query. still trying to figure out why the numbers don't check out

temp.graphql
xBalbinus: Got it - combining all the queries asap! Will have this in the morning.

xBalbinus: Hm, so groupBy is somehow not a thing: 
Screen Shot 2022-11-08 at 8 23 20 AM
Will indeed attempt the filter method though; I think I can recombine a lot of them into one.

xBalbinus: Apparently, groupBy is not possible.

As to how the aggregates work, I'm not sure if there is another way around besides for splitting them into individual aggregate queries like I did previously...

Though, summing the token send / receives into larger aggregates, total received is clearly greater than total sent (if that provides some insight).

Will add another query for a non-affected token as a sanity check.

xBalbinus:
Update:

Working around the limitations of groupBy, a new table was created
https://api.goldsky.io/c/nomad/gql/console/data/default/schema/public/views/affected_token_amounts/browse

Query method has been updated.

Unfortunately, numbers still don't work out and requires investigation as to why. Methods for creating the new table are according to the methodology described above.

ErinHales:
Rebase and it should fix the failing checks

xBalbinus:
Rebasing, tysm for the fix!

ErinHales (approved):
Double checked all the numbers, looks good to me!  Would wait for Anna to do a final review as well

prestwich:
dont merge main, rebase onto main 🫡

xBalbinus:
Hard resetted / modified event accountant afterwards

anna-carroll (approved):
Just reviewed everything against the latest numbers queried in Goldsky. All numbers match up. LGTM

PR #516, opened 2022-11-08
feat: Batch support for remote and local calls at the same time

odyslam:
- local and remote calls will be output to a different files, as nesting is not supported by current json output lib

### DIFF
--- a/packages/contracts-ops/contracts/CallBatch.sol
+++ b/packages/contracts-ops/contracts/CallBatch.sol
@@ -32,10 +32,9 @@ abstract contract CallBatch is Script {
         bool _overwrite
     ) public {
         require(localDomain == 0, "already initialized");
-        require(bytes(outputFile.path).length == 0, "already initialized");
+        require(_localDomain != 0, "can't pass zero domain");
         localDomainName = _localDomainName;
         localDomain = _localDomain;
-        outputFile.path = _outputFilePath;
         outputFile.overwrite = _overwrite;
         outputFile.path = string(
             abi.encodePacked("./actions/", _outputFilePath)
@@ -224,7 +223,6 @@ abstract contract CallBatch is Script {
 
     function writeCallBatch(bool recovery) public {
         require(localDomain != 0, "must initialize");
-        require(bytes(outputFile.path).length != 0, "must initialize");
         require(!written, "already written");
         // write raw local & remote calls to file
         JsonWriter.Buffer memory buffer = JsonWriter.newBuffer();
--- a/packages/contracts-ops/contracts/CallBatch.sol
+++ b/packages/contracts-ops/contracts/CallBatch.sol
@@ -117,29 +117,43 @@ abstract contract CallBatch is Script {
 
     function writeCallList(
         JsonWriter.Buffer memory buffer,
+        string memory indent,
         GovernanceMessage.Call[] storage calls
     ) private {
         for (uint32 i = 0; i < calls.length; i++) {
-            writeCall(buffer, " ", calls[i], i == calls.length - 1);
+            writeCall(
+                buffer,
+                indent.nextIndent(),
+                calls[i],
+                i == calls.length - 1
+            );
         }
     }
 
-    function writeLocal(JsonWriter.Buffer memory buffer) private {
-        buffer.writeArrayOpen("", "local");
-        writeCallList(buffer, localCalls);
-        buffer.writeArrayClose("", false);
+    function writeLocal(JsonWriter.Buffer memory buffer, string memory indent)
+        private
+    {
+        buffer.writeArrayOpen(indent, "local");
+        writeCallList(buffer, indent, localCalls);
+        buffer.writeArrayClose(indent, false);
     }
 
-    function writeRemotes(JsonWriter.Buffer memory buffer) private {
+    function writeRemotes(JsonWriter.Buffer memory buffer, string memory indent)
+        private
+    {
         if (remoteDomains.length == 0) return;
-        buffer.writeObjectOpen("", "remotes");
+        buffer.writeObjectOpen(indent, "remotes");
         for (uint256 j; j < remoteDomains.length; j++) {
-            buffer.writeArrayOpen(" ", vm.toString(uint256(remoteDomains[j])));
-            writeCallList(buffer, remoteCalls[remoteDomains[j]]);
+            string memory inner = indent.nextIndent();
+            buffer.writeArrayOpen(
+                inner,
+                vm.toString(uint256(remoteDomains[j]))
+            );
+            writeCallList(buffer, inner, remoteCalls[remoteDomains[j]]);
             bool terminal = j == remoteDomains.length - 1;
-            buffer.writeArrayClose(" ", terminal);
+            buffer.writeArrayClose(inner, terminal);
         }
-        buffer.writeObjectClose("", false);
+        buffer.writeObjectClose(indent, false);
     }
 
     function writeRecoveryData(
@@ -174,17 +188,19 @@ abstract contract CallBatch is Script {
         buffer.writeKv(indent, "data", vm.toString(data), true);
     }
 
-    function writeBuilt(JsonWriter.Buffer memory buffer, bool recovery)
-        private
-    {
-        buffer.writeObjectOpen("", "built");
-        string memory indent = " ";
+    function writeBuilt(
+        JsonWriter.Buffer memory buffer,
+        string memory indent,
+        bool recovery
+    ) private {
+        buffer.writeObjectOpen(indent, "built");
+        string memory inner = indent.nextIndent();
         if (recovery) {
             // write local domain built
             bool isLastDomain = remoteDomains.length == 0;
             writeRecoveryData(
                 buffer,
-                indent,
+                inner,
                 localDomain,
                 localCalls,
                 isLastDomain
@@ -194,16 +210,16 @@ abstract contract CallBatch is Script {
                 isLastDomain = j == remoteDomains.length - 1;
                 writeRecoveryData(
                     buffer,
-                    indent,
+                    inner,
                     remoteDomains[j],
                     remoteCalls[remoteDomains[j]],
                     isLastDomain
                 );
             }
         } else {
-            writeGovernorData(buffer, indent);
+            writeGovernorData(buffer, inner);
         }
-        buffer.writeObjectClose("", true);
+        buffer.writeObjectClose(indent, true);
     }
 
     GovernanceMessage.Call[][] private builtRemoteCalls;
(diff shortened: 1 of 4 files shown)


packages/contracts-ops/contracts/CallBatch.sol
```diff
@@ -57,55 +97,111 @@ abstract contract CallBatch is Script {
...
 
-    function writeCallList(JsonWriter.Buffer memory buffer) private {
-        string memory indent = "";
-        string memory inner = indent.nextIndent();
-        string memory innerer = inner.nextIndent();
-        buffer.writeLine(indent, "{");
-        buffer.writeArrayOpen(inner, domain);
-        for (uint32 i = 0; i < calls.length; i++) {
-            writeCall(buffer, innerer, calls[i], i == calls.length - 1);
+    function writeCallList(JsonWriter.Buffer memory buffer, bool local)
```
anna-carroll, line 100: this should be refactored - the code in `if(local)` is exactly copied with the code in the loop over `remoteDomains`. also passing a boolean as a parameter to switch on and have different function behavior is not readable

i think you can do something like (not writing in an IDE / with compiler so this might not be exactly right)
```
function writeCallList(JsonWriter.Buffer memory buffer, GovernanceMessage.Call[] storage calls) private {
        string memory indent = "";
        string memory inner = indent.nextIndent();
        string memory innerer = inner.nextIndent();
        buffer.writeLine(indent, "{");
        buffer.writeArrayOpen(inner, domain);
        for (uint32 i = 0; i < calls.length; i++) {
            writeCall(buffer, innerer, calls[i], i == calls.length - 1);
       }
        buffer.writeArrayClose(inner, true);
        buffer.writeLine(indent, "}");
}

function writeLocal(JsonWriter.Buffer memory buffer) {
       writeCallList(buffer, localCalls);
}

function writeRemotes(JsonWriter.Buffer memory buffer) {
       for (uint256 j; j < remoteDomains.length; j++) {
              writeCallList(buffer, remoteCalls[j]);
       }
}

```

packages/contracts-ops/contracts/CallBatch.sol
```diff
@@ -57,55 +97,111 @@ abstract contract CallBatch is Script {
...
-        require(bytes(domain).length != 0, "must initialize");
-        require(bytes(outputFile.path).length != 0, "must initialize");
+        require(bytes(localDomain).length != 0, "must initialize");
+        require(bytes(outputFileLocal.path).length != 0, "must initialize");
         require(!complete, "already written");
         complete = true;
         JsonWriter.Buffer memory buffer = JsonWriter.newBuffer();
-        writeCallList(buffer);
-        buffer.flushTo(outputFile);
+        writeCallList(buffer, true);
```
anna-carroll, line 146: should we flush each domain's governance calls to individual files?

anna-carroll:
```
./actions/governance-ethereum.json
./actions/governance-avalanche.json
```
etc

odyslam: Depending on what we want. If we use vm.writeJson, we can create arbitrarily complex JSON structures quite easily. With the json library we have, we can't create anything nested, so the best we can do is flush them to different files.

anna-carroll: i think we definitely need to do that actually. as things stand i think the remote file wouldn't be actual JSON. and i don't think it would be clear which governance actions are for which domains

packages/contracts-ops/contracts/CallBatch.sol
```diff
@@ -57,55 +97,111 @@ abstract contract CallBatch is Script {
...
-        for (uint32 i = 0; i < calls.length; i++) {
-            writeCall(buffer, innerer, calls[i], i == calls.length - 1);
+    function writeCallList(JsonWriter.Buffer memory buffer, bool local)
+        private
+    {
+        if (local) {
+            string memory indent = "";
+            string memory inner = indent.nextIndent();
+            string memory innerer = inner.nextIndent();
+            buffer.writeLine(indent, "{");
```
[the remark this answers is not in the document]
prestwich: this is in addition to anna's above refactor

packages/contracts-ops/contracts/CallBatch.sol
prestwich: I think it should be 1 run, 1 output file. creating an unknown number or network-dependent number seems gross

packages/contracts-ops/contracts/CallBatch.sol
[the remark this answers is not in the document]
prestwich: I don't feel qualified to review vm.serializeJson usage

packages/contracts-ops/contracts/CallBatch.sol
```diff
@@ -11,37 +11,77 @@ import {TypeCasts} from "@nomad-xyz/contracts-core/contracts/XAppConnectionManag
 import {JsonWriter} from "./JsonWriter.sol";
 
 import "forge-std/Script.sol";
+import "forge-std/console2.sol";
```
anna-carroll, line 14: gratuitous?

odyslam:
Example output with finish

```
{
  "local": {
    "ethereum": [
      {
        "to": "0x000000000000000000000000000000000000000000000000000000000000beef",
        "data": "0x0101"
      }
    ]
  },
  "remote": {
    "evmos": [
      {
        "to": "0x0000000000000000000000000000000000000000000000000000000000beeeef",
        "data": "0x0101"
      }
    ],
    "moonbeam": [
      {
        "to": "0x0000000000000000000000000000000000000000000000000000000beeeeeeef",
        "data": "0x0101"
      }
    ]
  }
}
```

packages/contracts-ops/contracts/CallBatch.sol
```diff
@@ -16,32 +16,93 @@ abstract contract CallBatch is Script {
...
+    ) public {
+        require(bytes(localDomain).length == 0, "already initialized");
         require(bytes(outputFile.path).length == 0, "already initialized");
-        domain = _domain;
-        _outputFile = string(abi.encodePacked("./actions/", _outputFile));
-        outputFile.path = _outputFile;
+        outputFile.path = _outputFilePath;
         outputFile.overwrite = _overwrite;
+        localDomain = _localDomain;
+        remoteDomains = _remoteDomains;
```
anna-carroll, line 52: why is this passed at initialize? it should be dynamically updated when remote calls are pushed

odyslam: CallBatch doesn't inherit/use the config, so I assumed this was a design requirement.

odyslam: ah, I see what you mean.

packages/contracts-ops/contracts/CallBatch.sol
```diff
@@ -16,32 +16,91 @@ abstract contract CallBatch is Script {
     using JsonWriter for JsonWriter.Buffer;
     using JsonWriter for string;
 
-    GovernanceMessage.Call[] public calls;
+    string localDomainName;
```
anna-carroll, line 19: bit of a hack. the rub here is that 
1. we have been using strings for domains within forge scripts thus far
2. the `GovernanceRouter`, like other contracts in the solidity codebase, use uint32 for domains 
3. `CallBatch` doesn't have access to `Config`, and can't convert from the forge scripts' strings to the contracts' expected uint32

anna-carroll: we should probably reconcile this problem. a potential solution is overloading all of the config getters to work with uint32 or string domains. we can write a config function which converts from a uint32 domain to a string by looping over each of the networks and to find the matching string for a domain.

anna-carroll (approved):
have reviewed my work again this morning, looking good to me. i think the main thing i'd harp on is JSON writing sux right now, way too verbose and hard to read - but we can refactor and fix that in the future!

packages/contracts-ops/contracts/JsonWriter.sol
```diff
@@ -102,16 +102,23 @@ library JsonWriter {
         writeLine(buffer, indent, terminal ? "]" : "],");
     }
 
+    function writeObjectOpen(Buffer memory buffer, string memory indent)
```
prestwich, line 105: 👍

packages/contracts-ops/contracts/scripts/RotateUpdater.s.sol
```diff
@@ -52,38 +52,35 @@ contract RotateUpdaterLogic is Config, CallBatch {
     }
 }
 
-contract RotateUpdater is Script, RotateUpdaterLogic {
+abstract contract RotateUpdater is Script, RotateUpdaterLogic {
```
prestwich, line 55: why is this now abstract?

packages/contracts-ops/contracts/CallBatch.sol
prestwich: > we have been using strings for domains within forge scripts thus far

this is because the config lists networks by string, and has them keyed by string. i.e. our ops must have a string -> number mapping available to translate from config to solidity state

prestwich: >  a potential solution is overloading all of the config getters to work with uint32 or string domains.

sdk resolvedomain function

prestwich (changes requested):
should have been more explicit earlier, the ideal JSON output is a callbatch, which is a json format in the govern sdk

```

export interface Call {
  to: Address;
  data?: ethers.utils.BytesLike;
}

export interface RemoteContents {
  [domain: string]: Call[];
}

export interface CallBatchContents {
  local: Call[];
  remote: RemoteContents;
  built?: ethers.PopulatedTransaction;
}
```

packages/contracts-ops/contracts/scripts/RotateUpdater.s.sol
anna-carroll: to silence compiler warnings - as i said in slack, i thought it still allows you to run the script?

anna-carroll:
but the thing is, the sdk-govern needs an update itself to support building the callbatch for recovery

anna-carroll:
my proposed interface was: 

```
// same as before
export interface Call {
  to: Address;
  data?: ethers.utils.BytesLike;
}

// same as before
export interface RemoteContents {
  [domain: string]: Call[];
}

// delete built from CallBatchContents
export interface CallBatchContents {
  local: Call[];
  remote: RemoteContents;
}

// add GovernanceCallBatch which adds back built as it was before
export interface GovernanceCallBatch is CallBatchContents {
  built?: ethers.PopulatedTransaction;
}

// add RecoveryCallBatch with transactions built for each chain
export interface RecoveryBuilt {
  [domain: string]: ethers.PopulatedTransaction;
}

export interface RecoveryCallBatch is CallBatchContents {
  built?: RecoveryBuilt;
}
```

@prestwich

anna-carroll:
something here that makes this harder - we can't have domains as strings / names, we can't know the to for the router on each chain - is because the `CallBatch` doesn't have access to `Config`, as it does in `sdk-govern`. which is probably a cleaner separation of concerns, but makes mimicking the functionality harder

prestwich:
is there a reason that a recovery batch is not just a regular callbatch with no remote calls? Why differentiate?

anna-carroll:
the differentiation is only at the stage where you _build_ the CallBatch. before that, they're the same

prestwich:
then there's no reason to differentiate the types, no? just add a `buildRecovery()`?

anna-carroll:
the type of `built` is different for each one

anna-carroll:
perhaps the confusion is stemming from different intended usages in our minds for multichain scripts. if you're running a script which runs on all chains at once, and pushes governance calls for each chain, what's the intended behavior: 

- push all calls to the same CallBatch; build it differently 
- make a different CallBatch for each chain; built it for each chain as if it's the governor chain

odyslam:
@anna-carroll When the json cheatcodes mature just a bit, this will get **much** cleaner. It's a relatively easy refactor, assuming cheatcodes function as expected. 

I would suggest to merge this to unblock the upgradeCallBatches PR

PR #527, opened 2022-11-13
feat: deploy implementations for upgrade

anna-carroll:
## Motivation
To Upgrade the protocol, we must deploy fresh implementations of each contract before creating governance transactions

## Solution
Deploy implementations via forge script

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
+++ b/packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
@@ -28,9 +28,7 @@ abstract contract DeployImplementationsLogic is Script, Config {
     Home home;
     Replica replica;
     GovernanceRouter governanceRouter;
-    // type of BridgeRouter differs by domain,
-    // so store as address
-    address payable bridgeRouter;
+    BridgeRouter bridgeRouter;
     TokenRegistry tokenRegistry;
     BridgeToken bridgeToken;
 
@@ -62,11 +60,13 @@ abstract contract DeployImplementationsLogic is Script, Config {
         );
         // BridgeRouter
         if (keccak256(bytes(_domain)) == keccak256(bytes("ethereum"))) {
-            bridgeRouter = address(
-                new EthereumBridgeRouter(address(getAccountant(_domain)))
+            bridgeRouter = BridgeRouter(
+                address(
+                    new EthereumBridgeRouter(address(getAccountant(_domain)))
+                )
             );
         } else {
-            bridgeRouter = address(new BridgeRouter());
+            bridgeRouter = new BridgeRouter();
         }
         BridgeRouter(bridgeRouter).initialize(
             address(tokenRegistry),
@@ -133,7 +133,13 @@ contract DeployImplementations is DeployImplementationsLogic {
             address(governanceRouter),
             false
         );
-        writeImplementation(buffer, inner, "bridgeRouter", bridgeRouter, false);
+        writeImplementation(
+            buffer,
+            inner,
+            "bridgeRouter",
+            address(bridgeRouter),
+            false
+        );
         writeImplementation(
             buffer,
             inner,
--- a/packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
+++ b/packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
@@ -45,6 +45,7 @@ abstract contract DeployImplementationsLogic is Script, Config {
     function deployImplementations(string memory _domain) internal {
         // Home
         home = new Home(getDomainNumber(_domain));
+        // NOTE: must pass real value for UpdaterManager because it is queried
         home.initialize(IUpdaterManager(address(getUpdaterManager(_domain))));
         // Replica
         replica = new Replica(getDomainNumber(_domain));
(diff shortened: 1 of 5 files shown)


packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
```diff
@@ -0,0 +1,171 @@
...
+
+abstract contract DeployImplementationsLogic is Script, Config {
+    /*//////////////////////////////////////////////////////////////
+                          DEPLOYED CONTRACTS
+    //////////////////////////////////////////////////////////////*/
+
+    Home home;
+    Replica replica;
+    GovernanceRouter governanceRouter;
+    // type of BridgeRouter differs by domain,
```
prestwich, line 31: because EthBridgeRouter is a child contract, you can store it as a bridge router (until/unless you need to access ethbridgerouter-specific interface methods)

anna-carroll: https://github.com/nomad-xyz/monorepo/pull/527/commits/427d4b27f13c32836317cd54b5aea29cbcb5f951

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
```diff
@@ -0,0 +1,177 @@
...
+        replica = new Replica(getDomainNumber(_domain));
+        replica.initialize(0, address(0), bytes32(0), 0);
+        // GovernanceRouter
+        governanceRouter = new GovernanceRouter(
+            getDomainNumber(_domain),
+            getRecoveryTimelock(_domain)
+        );
+        governanceRouter.initialize(
+            address(getXAppConnectionManager(_domain)),
+            address(0)
```
prestwich, line 59: is this correct?

prestwich: oh. you're initializing the implementation? but with some real data?

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
```diff
@@ -0,0 +1,177 @@
...
+                address(
+                    new EthereumBridgeRouter(address(getAccountant(_domain)))
+                )
+            );
+        } else {
+            bridgeRouter = new BridgeRouter();
+        }
+        BridgeRouter(bridgeRouter).initialize(
+            address(tokenRegistry),
+            address(getXAppConnectionManager(_domain))
```
prestwich, line 73: why initialize implemetnation with real data?

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
```diff
@@ -0,0 +1,177 @@
...
+        );
+        // TokenRegistry
+        tokenRegistry = new TokenRegistry();
+        tokenRegistry.initialize(
+            address(bridgeToken),
+            address(getXAppConnectionManager(_domain))
```
prestwich, line 79: "

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
anna-carroll: the `xAppConnectionManager` is queried, so have to pass a real address there, but the second address (`recoveryManager`) is not queried, so it can be whatever. since it's the implementation, and this is `initialize` / not `constructor`, values don't matter as long as the tx doesn't revert

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
anna-carroll: i believe both of these would revert if you passed `address(0)` because they are cast as contracts in the `initialize` function

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
anna-carroll: the first param could be changed to `address(0)` but the second could not, i believe. i'll push a change for the first.

odyslam:
The original pr works and it just needs the latest forge-std to compile

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
anna-carroll: https://github.com/nomad-xyz/monorepo/pull/527/commits/ee6232b67f40972eefbcb7e7900b131aca5ff145

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
anna-carroll: https://github.com/nomad-xyz/monorepo/pull/527/commits/ee6232b67f40972eefbcb7e7900b131aca5ff145

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
anna-carroll: https://github.com/nomad-xyz/monorepo/pull/527/commits/ee6232b67f40972eefbcb7e7900b131aca5ff145

anna-carroll:
@prestwich i removed `initialize` and tested the script again on testnet - ready for re-review

prestwich (approved):

PR #529, opened 2022-11-13
feat: deploy option which adds initialize

anna-carroll:
## Motivation
Want to reduce potential messy chain state & false alarm bug reports by initializing implementations

## Solution
Add initialize implementations on top of existing deploy script

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
+++ b/packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
@@ -88,7 +88,6 @@ contract DeployImplementations is DeployImplementationsLogic {
         vm.startBroadcast();
         deployImplementations(_domain);
         write(buffer, indent.nextIndent(), _domain);
-        vm.stopBroadcast();
         // write final output
         buffer.writeObjectClose(indent, true);
         buffer.flushTo(outputFile);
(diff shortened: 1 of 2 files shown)


packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
```diff
@@ -88,7 +88,6 @@ contract DeployImplementations is DeployImplementationsLogic {
         vm.startBroadcast();
         deployImplementations(_domain);
         write(buffer, indent.nextIndent(), _domain);
-        vm.stopBroadcast();
```
prestwich, line 91: reasoning?

prestwich:
code ACK. the stop broadcast is to continue the broadcast into the init block if any?

packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
anna-carroll: foundry quirk (read: bug) - calling vm`.startBroadcast()` again in the parent script was reverting, whereas if you just never `stopBroadcast()` then it's fine

prestwich (approved):

PR #530, opened 2022-11-14
Feat: Upgrade Callbatches

odyslam:
## Motivation

To realise an upgrade (and update the implementation address of the beacons) we need to execute some governance call batches, 

## Solution

Build a script that reads an **updated** configuration file (meaning that it has the new implementations) and produce the required call batches in a JSON file

## PR Checklist

- [ ] Added Tests
- [x] Updated Documentation

### DIFF
--- a/packages/contracts-ops/contracts/Config.sol
+++ b/packages/contracts-ops/contracts/Config.sol
@@ -18,6 +18,7 @@ import {INomadProtocol} from "./test/utils/INomadProtocol.sol";
 
 import "forge-std/Vm.sol";
 import "forge-std/Test.sol";
+import "forge-std/console2.sol";
 
 abstract contract Config is INomadProtocol {
     Vm private constant vm =
--- a/packages/contracts-ops/contracts/Config.sol
+++ b/packages/contracts-ops/contracts/Config.sol
@@ -18,7 +18,6 @@ import {INomadProtocol} from "./test/utils/INomadProtocol.sol";
 
 import "forge-std/Vm.sol";
 import "forge-std/Test.sol";
-import "forge-std/console2.sol";
 
 abstract contract Config is INomadProtocol {
     Vm private constant vm =
--- a/packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
+++ b/packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
@@ -18,28 +18,6 @@ import {UpgradeBeaconController} from "@nomad-xyz/contracts-core/contracts/upgra
 
 contract UpgradeCallBatches is Script, Config, CallBatch {
     /*//////////////////////////////////////////////////////////////
-                                 BEACONS
-    //////////////////////////////////////////////////////////////*/
-
-    address homeBeacon;
-    address replicaBeacon;
-    address governanceRouterBeacon;
-    address bridgeRouterBeacon;
-    address tokenRegistryBeacon;
-    address bridgeTokenBeacon;
-
-    /*//////////////////////////////////////////////////////////////
-                             IMPLEMENTATIONS
-    //////////////////////////////////////////////////////////////*/
-
-    address homeImpl;
-    address replicaImpl;
-    address governanceRouterImpl;
-    address bridgeTokenImpl;
-    address tokenRegistryImpl;
-    address bridgeRouterImpl;
-
-    /*//////////////////////////////////////////////////////////////
                             GOVERNANCE CALLS
     //////////////////////////////////////////////////////////////*/
 
@@ -50,14 +28,8 @@ contract UpgradeCallBatches is Script, Config, CallBatch {
     bytes upgradeTokenRegistry;
     bytes upgradeBridgeToken;
 
-    address beaconController;
-    address governanceRouterProxy;
-
-    bytes executeCallBatchCall;
-
     string configFile;
     string[] domainNames;
-    string[] networksArray;
 
     function printCallBatches(
         string memory _configFile,
@@ -85,7 +57,6 @@ contract UpgradeCallBatches is Script, Config, CallBatch {
             outputFile,
             true
         );
-        networksArray = getNetworks();
         console2.log("Governance Actions have been output to", outputFile);
     }
 
@@ -96,9 +67,22 @@ contract UpgradeCallBatches is Script, Config, CallBatch {
     function generateGovernanceCalls(string memory domain) internal {
         upgradeHome = abi.encodeWithSelector(
             UpgradeBeaconController.upgrade.selector,
-            homeBeacon,
-            homeImpl
+            address(homeUpgrade(domain).beacon),
+            address(homeUpgrade(domain).implementation)
         );
+        string[] memory networks = getNetworks();
+        address replicaImpl;
+        address replicaBeacon;
+        for (uint256 i; i < networks.length; i++) {
+            if (getDomainNumber(networks[i]) != getDomainNumber(domain)) {
+                replicaImpl = address(
+                    replicaOfUpgrade(domain, networks[i]).implementation
+                );
+                replicaBeacon = address(
+                    replicaOfUpgrade(domain, networks[i]).beacon
+                );
+            }
+        }
         upgradeReplica = abi.encodeWithSelector(
             UpgradeBeaconController.upgrade.selector,
             replicaBeacon,
@@ -106,25 +90,26 @@ contract UpgradeCallBatches is Script, Config, CallBatch {
         );
         upgradeGovRouter = abi.encodeWithSelector(
             UpgradeBeaconController.upgrade.selector,
-            governanceRouterBeacon,
-            governanceRouterImpl
+            address(governanceRouterUpgrade(domain).beacon),
+            address(governanceRouterUpgrade(domain).implementation)
         );
         upgradeBridgeRouter = abi.encodeWithSelector(
             UpgradeBeaconController.upgrade.selector,
-            bridgeRouterBeacon,
-            bridgeRouterImpl
+            address(bridgeRouterUpgrade(domain).beacon),
+            address(bridgeRouterUpgrade(domain).implementation)
         );
         upgradeTokenRegistry = abi.encodeWithSelector(
             UpgradeBeaconController.upgrade.selector,
-            tokenRegistryBeacon,
-            tokenRegistryImpl
+            address(tokenRegistryUpgrade(domain).beacon),
+            address(tokenRegistryUpgrade(domain).implementation)
         );
         upgradeBridgeToken = abi.encodeWithSelector(
             UpgradeBeaconController.upgrade.selector,
-            bridgeTokenBeacon,
-            bridgeTokenImpl
+            address(bridgeTokenUpgrade(domain).beacon),
+            address(bridgeTokenUpgrade(domain).implementation)
         );
         uint32 domainNumber = getDomainNumber(domain);
+        address beaconController = address(getUpgradeBeaconController(domain));
         if (domainNumber == getDomainNumber(localDomainName)) {
             pushLocal(beaconController, upgradeBridgeToken);
             pushLocal(beaconController, upgradeTokenRegistry);
--- a/packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
+++ b/packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
@@ -17,20 +17,6 @@ import {CallBatch} from "../CallBatch.sol";
 import {UpgradeBeaconController} from "@nomad-xyz/contracts-core/contracts/upgrade/UpgradeBeaconController.sol";
 
 contract UpgradeCallBatches is Script, Config, CallBatch {
-    /*//////////////////////////////////////////////////////////////
-                            GOVERNANCE CALLS
-    //////////////////////////////////////////////////////////////*/
-
-    bytes upgradeHome;
-    bytes upgradeReplica;
-    bytes upgradeGovRouter;
-    bytes upgradeBridgeRouter;
-    bytes upgradeTokenRegistry;
-    bytes upgradeBridgeToken;
-
-    string configFile;
-    string[] domainNames;
-
     function printCallBatches(
         string memory _configFile,
         string[] memory _domainNames,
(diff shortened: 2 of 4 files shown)


prestwich:
please fill out the motivation/solution sections

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,236 @@
...
+        __CallBatch_initialize(
+            localDomainName,
+            getDomainNumber(localDomainName),
+            outputFile,
+            true
+        );
+        networksArray = getNetworks();
+    }
+
+    function loadBeacons(string memory domain) internal {
```
prestwich, line 92: why is copying to local state preferable to reading from the config via getters?

i.e. why are we doing 


```
address a;

load() {
   a = a();
}

do_thing() {
   a;
}
```

instead of 
```
do_thing() {
  a()
}
```

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,236 @@
...
+            "Domain Number: ",
+            vm.toString(uint256(getDomainNumber(domain)))
+        );
+        console2.log("Function signature: upgrade(address, address)");
+        console2.log(
+            "Arguments: <contract_beacon>, <new_implementation_address>"
+        );
+
+        upgradeHome = abi.encodeWithSignature(
+            "upgrade(address,address)",
```
prestwich, line 148: `encodeWithSignature` is generally less preferable to `encodeWithSelector` as the latter is less brittle to interface changes

packages/contracts-ops/contracts/Config.sol
```diff
@@ -18,6 +18,7 @@ import {INomadProtocol} from "./test/utils/INomadProtocol.sol";
 
 import "forge-std/Vm.sol";
 import "forge-std/Test.sol";
+import "forge-std/console2.sol";
```
prestwich, line 21: leftover?

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
anna-carroll: agree it's cleaner to read from config imo

if we're doing multiple scripts at a time i don't want stale local state from a diff chain causing issues (also more stateful for the developer to think about while reading and editing the code, which is not preferable)

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,236 @@
...
+
+        upgradeBridgeToken = abi.encodeWithSignature(
+            "upgrade(address,address)",
+            bridgeTokenBeacon,
+            bridgeTokenImpl
+        );
+        console2.log("Upgrade Bridge Token");
+        console2.logBytes(upgradeBridgeToken);
+        if (compareStrings(domain, localDomainName)) {
+            pushLocal(beaconController, upgradeBridgeToken);
```
anna-carroll, line 194: tbh this is a refactor i was thinking would make sense for CallBatch - just `pushCall` which decides whether the call is local or remote

prestwich: the local/remote in callbatch is really referring to governing/non-governing in most cases 🤔

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,236 @@
...
+    address beaconController;
+    address governanceRouterProxy;
+
+    bytes executeCallBatchCall;
+
+    string configFile;
+    string[] domainNames;
+    string[] networksArray;
+
+    function run(
```
anna-carroll, line 60: more informative fn name would be ideal

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,236 @@
...
+                       GOVERNANCE CALL GENERATORS
+    //////////////////////////////////////////////////////////////*/
+
+    function generateGovernanceCalls(string memory domain) internal {
+        title("BeaconController upgrade encoded calls for", domain);
+        console2.log(
+            "Domain Number: ",
+            vm.toString(uint256(getDomainNumber(domain)))
+        );
+        console2.log("Function signature: upgrade(address, address)");
```
anna-carroll, line 142: for synchronous functions which execute in seconds (or less), i think it's not super helpful to do a bunch of console.logging. 

for async scripts like deployments or otherwise sending transactions, logging is useful, but foundry has very informative logging built-in, so i also don't find it that helpful in that case

i feel like adding inline comments to the code is better (and easier to read) and the person running the script just knows that it's doing what the code does

anna-carroll (commented):
💬

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
odyslam: This is a best practice for forge scripts that we try to make a standard, that is `run` is the entrypoint. I would prefer to keep it, but I can change it if y'all insist

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
odyslam: It's helpful to visually see the result of the script with the calldata of every batch nicely printed. Far easier than reading the json file alone. 

I can remove them, but personally they seem helpful.

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,236 @@
...
+        console2.logBytes(upgradeTokenRegistry);
+
+        upgradeBridgeToken = abi.encodeWithSignature(
+            "upgrade(address,address)",
+            bridgeTokenBeacon,
+            bridgeTokenImpl
+        );
+        console2.log("Upgrade Bridge Token");
+        console2.logBytes(upgradeBridgeToken);
+        if (compareStrings(domain, localDomainName)) {
```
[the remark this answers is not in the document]
odyslam: sounds good, will remove

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,236 @@
...
+        console2.log(title1, title2);
+        console2.log("===========================");
+        console2.log(" ");
+    }
+
+    function compareStrings(string memory a, string memory b)
+        internal
+        returns (bool)
+    {
+        return (keccak256(abi.encodePacked(a)) ==
```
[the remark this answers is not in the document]
odyslam: good observation

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
odyslam: okay, changed

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
anna-carroll: well, in this case, most of what's being logged is
- each domain (not much easier to read than JSON keys)
- a hard-coded reminder of the function signature & argument structure (we know this already) 
- the raw bytes (not really easier to read than JSON)

anna-carroll: maybe if it was something like 

```
"Home Beacon:", homeBeacon
"Home New Implementation:", homeImpl
```

then that would indeed be easier to read than JSON.

but again, what is the main purpose of logging that type of thing? it feels like it would be useful mainly for debugging while you're developing, to run the script and make sure it's hitting the domains you expect and the arguments you expect. 

but i would rather that we ensure that the script actually does what we expect via 
- robust code review 
- fork testing
- staging testing 
- maybe spot checking JSON a couple times until we gain confidence in the script

than by utilizing a solution that makes the code itself harder to read (imo)

odyslam: Gotcha. Removing

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,187 @@
...
+            homeBeacon,
+            homeImpl
+        );
+        console2.log("Upgrade Home");
+        console2.logBytes(upgradeHome);
+
+        upgradeReplica = abi.encodeWithSelector(
+            UpgradeBeaconController.upgrade.selector,
+            replicaBeacon,
+            replicaImpl
```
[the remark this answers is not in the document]
odyslam: oooof,

packages/contracts-ops/contracts/Config.sol
anna-carroll: @odyslam why convo resolved - shouldn't this be removed?

odyslam: Yes, mistook the console2 import for the main script

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,129 @@
...
+
+    function printCallBatches(
+        string memory _configFile,
+        string[] memory _domainNames,
+        string memory _localDomainName,
+        bool recovery
+    ) external {
+        localDomainName = _localDomainName;
+        domainNames = _domainNames;
+        configFile = _configFile;
```
anna-carroll, line 42: nit: no need to put this in storage. just call `__Config_initialize` here

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,129 @@
...
+    string[] domainNames;
+
+    function printCallBatches(
+        string memory _configFile,
+        string[] memory _domainNames,
+        string memory _localDomainName,
+        bool recovery
+    ) external {
+        localDomainName = _localDomainName;
+        domainNames = _domainNames;
```
anna-carroll, line 41: nit: why put this in storage?

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,129 @@
...
+        );
+        console2.log("Governance Actions have been output to", outputFile);
+    }
+
+    /*//////////////////////////////////////////////////////////////
+                       GOVERNANCE CALL GENERATORS
+    //////////////////////////////////////////////////////////////*/
+
+    function generateGovernanceCalls(string memory domain) internal {
+        upgradeHome = abi.encodeWithSelector(
```
anna-carroll, line 68: why use storage variables for the calldata instead of local memory?

anna-carroll (commented):
comin' together. a couple nits about storage vs. memory. i think they're non-blocking if you prefer this stylistically. but i don't like storage variables that might persist incorrectly across different domains

taking one last pass at logic

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,129 @@
...
+            UpgradeBeaconController.upgrade.selector,
+            address(homeUpgrade(domain).beacon),
+            address(homeUpgrade(domain).implementation)
+        );
+        string[] memory networks = getNetworks();
```
anna-carroll, line 64: instead of `getNetworks`, you can use `getConnections(domain)`. then, you can remove the `if()` statement

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,129 @@
...
+        address replicaImpl;
+        address replicaBeacon;
+        for (uint256 i; i < networks.length; i++) {
```
anna-carroll, line 67: you only need to submit 1 call to upgrade all replicas. the loop is misleading / uneccessary. 
you can just do 
```
replicaOfUpgrade(domain, getConnections(domain)[0]).implementation
```

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -0,0 +1,129 @@
...
+            if (getDomainNumber(networks[i]) != getDomainNumber(domain)) {
+                replicaImpl = address(
+                    replicaOfUpgrade(domain, networks[i]).implementation
+                );
+                replicaBeacon = address(
+                    replicaOfUpgrade(domain, networks[i]).beacon
+                );
+            }
+        }
+        upgradeReplica = abi.encodeWithSelector(
```
anna-carroll, line 77: suggested refactor: 

```
function pushUpgrade(Upgrade _upgrade, string memory domain) private {
        bytes call = abi.encodeWithSelector(
            UpgradeBeaconController.upgrade.selector,
            address(_upgrade.beacon),
            address(_upgrade.implementation)
        );
        uint32 domainNumber = getDomainNumber(domain);
        address beaconController = address(getUpgradeBeaconController(domain));
        if (domainNumber == getDomainNumber(localDomainName)) {
             pushLocal(beaconController, call);
        } else {
             pushRemote(beaconController, call, domainNumber);
        }       
}
```

odyslam:
@anna-carroll removed all storage vars

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
anna-carroll:
```
pushUpgrade(homeUpgrade(domain), domain);
pushUpgrade(replicaOfUpgrade(domain, getConnections(domain)[0]), domain);
```
... etc

anna-carroll: note this would be a pretty valid place to store domain to storage, to avoid it being passed as a parameter redundantly all the time. just need to make very sure you are actually updating that storage variable for each loop.

odyslam: This is a very good suggestion. Thanks

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
odyslam: ah, didn't in fact know of that field.

anna-carroll (approved):
very clean!! love it

anna-carroll:
some thoughts: 
- would be nice if this was written in a way that we could layer this on top of the current deploy implementations script, per some of @prestwich thoughts - the current workflow requires all of the deployments to finish, then update config, then run this script
- it would be a better workflow if the deploy implementations script could edit the config in-place, and then scripts like this could be layered on top of that and read from the updated config
- this is another point in favor of logic contracts operating on a single-chain basis; the entrypoint / script runner can loop through all chains, but it's easier to layer functionality on top of each other if the logic operates on one domain at a time 

I feel like a `Config.sol` refactor that allows us to 
   - read config from file
   - edit config in-place
   - write the full values of an edited config to file 
would be super OP to make these scripts better to work with

prestwich:
i generally dont apparove of updating a config in place. copy to new
location and update at new location

odyslam:
We can take some of the functionality from the multi-chain implementation deployment that replaced the configuration in-place. The ideal addition would be to first copy the configuration and the replace in-place.

anna-carroll:
> i generally dont apparove of updating a config in place

what if we didn't update the config file in-place, but the values in memory - then write the completed file at a new location at the end of the script

PR #531, opened 2022-11-14
fix: properly formatted tokenId

odyslam:
## Motivation
(from slack, by James)
- the input has domain == localDomain
- the input id has non-empty top bytes, so it passes the vm.assume
- the input id has empty bottom 20 bytes, so getLocalAddress shows the local address as 0
- this triggers the if block in ensureLocalToken, and deploys a new registry token
- ensureLocalToken returns the new token, which is not equal to the lower 20 bytes of the input
this is a condition of use violation. The token registry assumes that there is no local token at address(0), and therefore, the code assumes that it’ll never get a local token ID with empty bottom 20 bytes

## Solution

Ensure in the test that the ID is formatted correctly, which means that the lower 20 bytes should be != 0

### DIFF
--- a/packages/contracts-bridge/contracts/test/TokenRegistry.t.sol
+++ b/packages/contracts-bridge/contracts/test/TokenRegistry.t.sol
@@ -128,7 +128,7 @@ contract TokenRegistryTest is BridgeTest {
         public
     {
         vm.assume(domain != 0 && id != bytes32(0));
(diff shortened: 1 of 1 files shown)


prestwich (changes requested):
NACK. diagnosis is incorrect, see slack.

prestwich:
please use the solution & motivation blocks in the PR template/

odyslam:
ack

prestwich (approved):

PR #534, opened 2022-11-14
feat: replace in-place the config implementation values

odyslam:
## Motivation

We don't want to manually update the config with the new implementation addresses

## Solution

The upgrade script will update in-place the config file. After every invocation, the config file will update its implementation addresses for the domain that just got deployed.

## PR Checklist

- [x] Added Tests
- [x] Updated Documentation

### DIFF
--- a/packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
+++ b/packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
@@ -64,92 +64,80 @@ abstract contract DeployImplementationsLogic is Script, Config {
 }
 
 contract DeployImplementations is DeployImplementationsLogic {
-    using JsonWriter for JsonWriter.Buffer;
-    using JsonWriter for string;
-
-    JsonWriter.File outputFile;
-
     // entrypoint
-    function deploy(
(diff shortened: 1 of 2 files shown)


packages/contracts-ops/contracts/scripts/DeployImplementations.s.sol
```diff
@@ -64,92 +64,80 @@ abstract contract DeployImplementationsLogic is Script, Config {
...
-        writeImplementation(
-            buffer,
-            inner,
-            "governanceRouter",
-            address(governanceRouter),
-            false
+    function updateImpl(string memory _domain, string memory _configPath)
+        internal
+    {
+        string memory valueKey = string(
```
anna-carroll, line 79: this will read a bit nicer if you use the utils in the `Config.sol` for generating paths in the config :)

anna-carroll: then you can just pass that directly in vm.writeJson as a param, instead of overwriting the same memory variable repeatedly

anna-carroll (approved):

PR #536, opened 2022-11-17
fix reboot for script running

anna-carroll:
## Motivation
- without vm.broadcast, transactions don't send
- using `call` instead of `staticcall` was sending 6 transactions on-chain

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
+++ b/packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
@@ -41,7 +41,7 @@ contract UpgradeCallBatchLogic is Script, Config, CallBatch {
         string memory contractName
     ) private {
         // check if upgrade is unnecessary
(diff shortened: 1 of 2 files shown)


prestwich (approved):

packages/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol
```diff
@@ -41,7 +41,7 @@ contract UpgradeCallBatchLogic is Script, Config, CallBatch {
         string memory contractName
     ) private {
         // check if upgrade is unnecessary
-        (, bytes memory result) = address(_upgrade.beacon).call("");
+        (, bytes memory result) = address(_upgrade.beacon).staticcall("");
```
odyslam, line 44: We can do 
```solidity
address _current = vm.load(address(_upgrade.beacon), bytes32("1")));
```

PR #538, opened 2022-11-18
Feat: CoreForkBase

odyslam:
This is a draft PR, and it doesn't compile. 

Before investing a bunch of time in refactoring, this is a blueprint of how I imagine the test suite to be modular:

- Tests
- Protocol
- Environment

Let's start from the ground up:
- Environment: It sets the environment where the tests will run. This can be one of: [Mock, Local, Fork]. It creates the contracts that the tests access.
- Protocol: This is a very thin layer and serves as an API between the Tests and Environment. Basically it informs the tests on what contracts are made available by the Environment. In the current PR this layer is integrated into the Environment.
- Test: The actual tests (e.g ReplicaTest).

With this design, we have the following contract inheritance:|
- FORK: ForCore --> <CHAIN><ENV>ForkBase --> Replica Test
- LOCAL: LocalCore (refactor of NomadBase.sol) --> Replcia Test
- MOCK: MockCore -> Replica Test

I have added Mock because it's helpful to test some system parts in isolation. For example, Bridge tests might be Local
(LocalBridge) for the bridge contracts but MockCore for the Core contracts.

Before diving into refactoring NomadBase to be able to properly inherit ReplicaTest to a ForkReplicaTest, I want to ask for some feedback. I spent some time thinking about how I should go about this, and this approach seems the sanest -- assuming we want modularity between the systems (fork, local, etc.) and not build them in isolation.

For the actual inheritance, I imagine all the test contracts to be abstract and then have:
- ReplicaTest as abstract
- LocalReplicaTest is ReplicaTest, LocalCore
- EthProdReplicaForkTest is ReplicaTest, EthProdCoreFork

### DIFF
--- a/packages/contracts-core/contracts/test/fork/ForkBase.sol
+++ b/packages/contracts-core/contracts/test/fork/ForkBase.sol
@@ -1,0 +1,81 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity 0.7.6;
+pragma abicoder v2;
+
+import "forge-std/Test.sol";
+import {UpgradeCallBatchLogic} from "@nomad-xyz/contracts-ops/contracts/scripts/UpgradeCallBatches.s.sol";
+import {Home} from "../harnesses/HomeHarness.sol";
+import {ReplicaHarness} from "../harnesses/ReplicaHarness.sol";
+import {GovernanceRouterHarness} from "../harnesses/GovernanceRouterHarness.sol";
+import {XAppConnectionManager} from "../../XAppConnectionManager.sol";
+import {UpdaterManager} from "../../UpdaterManager.sol";
+
+contract ForkBase is UpgradeCallBatchLogic, Test {
+    Home home;
+    ReplicaHarness replica;
+    GovernanceRouterHarness governanceRouter;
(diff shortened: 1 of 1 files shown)


packages/contracts-core/contracts/test/fork/ForkBase.sol
```diff
@@ -0,0 +1,81 @@
...
+    UpdaterManager updaterManager;
+
+    string configName;
+    string govDomain;
+
+    string local;
+    string remote;
+    uint32 remoteDomain;
+    // Defined in CallBatch, which is inherited via UpgradeCallBatchLogic
+    // uint32 localDomain;
```
prestwich, line 27: this has created conflicts with bridge router tests as well. we should consider renaming callbatch's to `batchLocalDomain` ?

odyslam: LGTM

anna-carroll:
I think we will refactor to something along these lines very soon, esp once foundry race condition for fork testing is solid

anna-carroll:
I think we will refactor to something along these lines very soon, esp once foundry race condition for fork testing is solid. For now, closing the PR as i believe it was mostly meant to be illustrative

PR #540, opened 2022-11-18
feat: GovernanceRouter unit tests in fork mode

odyslam:
## Motivation

Test the GovernanceRouter with on-chain state

## Solution

Refactor (a bit) the unit tests and inherit them to the fork environment

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-ops/contracts/test/Reboot.t.sol
+++ b/packages/contracts-ops/contracts/test/Reboot.t.sol
@@ -73,19 +73,21 @@ contract RebootTest is
         optimisticTimeout = replica.optimisticSeconds();
         // setup fake app handlers
         setUpBadHandlers();
-        // Upgrade to Harness
-        // deploy ReplicaHarness implementation
+        // upgrade contracts to harness
+        upgradeToHarness();
+    }
+
+    function upgradeToHarness() public {
+        upgradeReplicaHarness();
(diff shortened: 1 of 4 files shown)


anna-carroll:
looks good i think. does governance router have any tests that interact with real home functionality?

packages/contracts-ops/contracts/test/Reboot.t.sol
```diff
@@ -27,56 +33,70 @@ contract RebootTest is
             NomadTestWithUpdaterManager,
             ReplicaHandlers,
             ReplicaTest,
-            HomeTest
+            HomeTest,
+            GovernanceRouterTest
         )
     {
+        // ALL
         vm.createSelectFork(vm.envString("RPC_URL"), 15_977_625);
```
prestwich, line 41: This function is too large, I can't tell what it's doing. Break it up into discrete steps

prestwich: rule of thumb: anything that invokes writejson should be in a different scope that pairs it with the thing that generated whatever it is that is being written

anna-carroll:
refactored: https://github.com/nomad-xyz/monorepo/pull/544

PR #543, opened 2022-11-19
refactor: fork tests in own files

anna-carroll:
## Motivation
Race condition in foundry spinning up fork tests was fixed; tests can now be refactored more cleanly into their own files

## Solution
Do the thing :)

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-ops/contracts/test/RebootReplica.t.sol
+++ b/packages/contracts-ops/contracts/test/RebootReplica.t.sol
@@ -1,0 +1,52 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity 0.7.6;
+pragma abicoder v2;
+
+import "forge-std/Test.sol";
+import {RebootTest} from "./Reboot.t.sol";
+import {NomadTest} from "@nomad-xyz/contracts-core/contracts/test/utils/NomadTest.sol";
+import {ReplicaHarness} from "@nomad-xyz/contracts-core/contracts/test/harnesses/ReplicaHarness.sol";
+import {ReplicaTest} from "@nomad-xyz/contracts-core/contracts/test/Replica.t.sol";
+
+contract ReplicaRebootTest is RebootTest, ReplicaTest {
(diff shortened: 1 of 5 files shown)


packages/contracts-ops/contracts/test/RebootReplica.t.sol
```diff
@@ -0,0 +1,52 @@
...
+        pushSingleUpgrade(replicaOfUpgrade(localDomainName, remote));
+        prankExecuteRecoveryManager(
+            address(getGovernanceRouter(localDomainName)),
+            getDomainNumber(localDomainName)
+        );
+    }
+
+    // REPLICA
+    function test_acceptLeafCorrectProof() public override {
+        // TODO: test proof somehow?
```
prestwich, line 50: use one from the s3 bucket?

packages/contracts-ops/contracts/test/RebootReplica.t.sol
```diff
@@ -0,0 +1,52 @@
...
+        reloadConfig();
+        pushSingleUpgrade(replicaOfUpgrade(localDomainName, remote));
+        prankExecuteRecoveryManager(
+            address(getGovernanceRouter(localDomainName)),
+            getDomainNumber(localDomainName)
+        );
+    }
+
+    // REPLICA
+    function test_acceptLeafCorrectProof() public override {
```
prestwich, line 49: side note, rather than overriding with empty body

```
modifier skipForking() {
    try vm.activeFork() {
        return;
    } catch (bytes memory) {
        _;
    }
}
```

prestwich (approved):

packages/contracts-ops/contracts/test/RebootReplica.t.sol
anna-carroll: do we want to introduce a dependency on the processor agents / S3 in our fork testing? 

i am open to it, but assumed the answer was no

PR #545, opened 2022-11-20
tweaks to home and replica fork tests

anna-carroll:
…ected

## Motivation
Want to ensure fork tests are being set up and run as expected

## Solution
Remove some unnecessary state vars as there are many inherited contracts. 
Add some assertions as guard rails.

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-ops/contracts/test/RebootHome.t.sol
+++ b/packages/contracts-ops/contracts/test/RebootHome.t.sol
@@ -15,6 +15,8 @@ contract HomeRebootTest is RebootTest, HomeTest {
         home = HomeHarness(address(getHome(localDomainName)));
         upgradeHomeHarness();
         updaterManager = getUpdaterManager(localDomainName);
+        // check fork setUp
+        assertEq(home.updater(), updaterAddr);
     }
 
     function upgradeHomeHarness() public {
@@ -28,11 +30,16 @@ contract HomeRebootTest is RebootTest, HomeTest {
             coreAttributePath(localDomainName, "home.implementation")
         );
         reloadConfig();
-        pushSingleUpgrade(homeUpgrade(localDomainName));
+        pushSingleUpgrade(homeUpgrade(localDomainName), localDomainName);
         prankExecuteRecoveryManager(
             address(getGovernanceRouter(localDomainName)),
             getDomainNumber(localDomainName)
         );
+        // assert beacon has been upgraded to harness
+        (, bytes memory result) = address(homeUpgrade(localDomainName).beacon)
+            .staticcall("");
+        address _current = abi.decode(result, (address));
+        assertEq(_current, address(homeHarnessImpl));
     }
 
     //////////////////////// HOME ////////////////////////
--- a/packages/contracts-ops/contracts/test/RebootHome.t.sol
+++ b/packages/contracts-ops/contracts/test/RebootHome.t.sol
@@ -9,23 +9,25 @@ import {HomeTest} from "@nomad-xyz/contracts-core/contracts/test/Home.t.sol";
 import {HomeHarness} from "@nomad-xyz/contracts-core/contracts/test/harnesses/HomeHarness.sol";
 
 contract HomeRebootTest is RebootTest, HomeTest {
+    address homeHarnessImpl;
+
     function setUp() public override(NomadTest, HomeTest) {
         setUpReboot(1, "home");
-        // HOME
+        // upgrade to harness
         home = HomeHarness(address(getHome(localDomainName)));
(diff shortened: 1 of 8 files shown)


packages/contracts-ops/contracts/test/RebootHome.t.sol
```diff
@@ -15,6 +15,8 @@ contract HomeRebootTest is RebootTest, HomeTest {
         home = HomeHarness(address(getHome(localDomainName)));
         upgradeHomeHarness();
         updaterManager = getUpdaterManager(localDomainName);
+        // check fork setUp
+        assertEq(home.updater(), updaterAddr);
```
prestwich, line 19: no assertions in `setUp`. due to a quirk in Foundry, these will not cause tests to fail.. instead add a `test_setUp` that makes this assertion

packages/contracts-ops/contracts/test/RebootHome.t.sol
```diff
@@ -28,11 +30,16 @@ contract HomeRebootTest is RebootTest, HomeTest {
...
         );
         reloadConfig();
-        pushSingleUpgrade(homeUpgrade(localDomainName));
+        pushSingleUpgrade(homeUpgrade(localDomainName), localDomainName);
         prankExecuteRecoveryManager(
             address(getGovernanceRouter(localDomainName)),
             getDomainNumber(localDomainName)
         );
+        // assert beacon has been upgraded to harness
+        (, bytes memory result) = address(homeUpgrade(localDomainName).beacon)
```
prestwich, line 39: should we make a convenience method for this on nomad test?

`getBeaconImplementation`

prestwich: `readBeacon`

packages/contracts-ops/contracts/test/RebootReplica.t.sol
```diff
@@ -15,11 +15,13 @@ contract ReplicaRebootTest is RebootTest, ReplicaTest {
...
             address(getReplicaOf(localDomainName, remote))
         );
-        assertEq(replica.updater(), updaterAddr);
         upgradeReplicaHarness();
+        // Replica test vars
         committedRoot = replica.committedRoot();
         optimisticTimeout = replica.optimisticSeconds();
         setUpBadHandlers();
+        // check fork setup
+        assertEq(replica.updater(), updaterAddr);
```
prestwich, line 24: no asserts in setup

packages/contracts-ops/contracts/test/RebootHome.t.sol
```diff
@@ -28,11 +30,16 @@ contract HomeRebootTest is RebootTest, HomeTest {
...
+            .staticcall("");
+        address _current = abi.decode(result, (address));
+        assertEq(_current, address(homeHarnessImpl));
```
prestwich, line 42: no asserts in methods called in setup. consider naming methods called in setup `setUp_doThing` so that we know which they are

packages/contracts-ops/contracts/test/RebootHome.t.sol
anna-carroll: oh weird, i thought that was a bug that they had fixed?

odyslam: I think it's fixed, but as a general pattern, we aligned that it's better to create a test function that doesn't do any state changes but tests the states of the setUp.

packages/contracts-ops/contracts/test/RebootHome.t.sol
anna-carroll: https://github.com/nomad-xyz/monorepo/pull/545/commits/a50e36eac21055df15997d24735ae1ce33d63b64

packages/contracts-ops/contracts/test/RebootReplica.t.sol
anna-carroll: https://github.com/nomad-xyz/monorepo/pull/545/commits/a50e36eac21055df15997d24735ae1ce33d63b64

prestwich (approved):

prestwich:
in the test, the sender is different. need to vm.prank the expected sender, or change the expectation

anna-carroll:
> in the test, the sender is different. need to vm.prank the expected sender, or change the expectation

this test doesn't appear to be following the pattern of other expectEmits for the home contract, changed it https://github.com/nomad-xyz/monorepo/pull/545/commits/ad37e37bb6efaeed9d5636853c6de175aede6b33

PR #550, opened 2022-11-22
fix: callbatch does not revert on empty batches

prestwich:
Indepedent, may be merged before or after any other PR

### DIFF
--- a/packages/contracts-ops/contracts/CallBatch.sol
+++ b/packages/contracts-ops/contracts/CallBatch.sol
@@ -70,10 +70,9 @@ abstract contract CallBatch {
 
     // only works if recovery is not active and localDomain is governor domain
     function prankExecuteGovernor(address router) public {
-        require(
-            localCalls.length != 0 || remoteDomains.length != 0,
-            "no calls pushed"
-        );
+        if (localCalls.length == 0 && remoteDomains.length == 0) {
+            return;
+        }
         // prank governor
         address gov = GovernanceRouter(router).governor();
         require(gov != address(0), "!gov chain");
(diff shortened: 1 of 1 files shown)


packages/contracts-ops/contracts/CallBatch.sol
```diff
@@ -70,10 +70,9 @@ abstract contract CallBatch {
 
     // only works if recovery is not active and localDomain is governor domain
     function prankExecuteGovernor(address router) public {
-        require(
-            localCalls.length != 0 || remoteDomains.length != 0,
-            "no calls pushed"
-        );
+        if (localCalls.length == 0 && remoteDomains.length == 0) {
```
anna-carroll, line 73: hmmm why though? i added this because when i was developing fork tests i was inadvertently not pushing to the batch right, executing, and confused why the effects didn't take. it seems helpful for the developer to know if they're executing no-op batches?

prestwich: because the fork test reboot upgrade code will sometimes push nothing, and then attempt to execute it

anna-carroll: yeah, that's the counterpoint i considered too, but actually they will always deploy the latest local version of implementations & upgrade them. 

that would change tho if we could implement the bytecode check that you mentioned, so i do take your point

anna-carroll: do you think still worth it, even tho it won't cause reboot to fail / may hide developer errors during usage? i am sorta tossup

prestwich: > because the fork test reboot upgrade code will sometimes push nothing, and then attempt to execute it

I made this change because the revery was blocking bridgerouter progress. i.e. the change is because I encountered this issue

anna-carroll (approved):

PR #552, opened 2022-11-23
feat(contracts-ops): tokenRegistry fork tests

odyslam:
## PR Checklist

- [x] Added Tests
- [x] Updated Documentation

### DIFF
--- a/packages/contracts-bridge/contracts/test/TokenRegistry.t.sol
+++ b/packages/contracts-bridge/contracts/test/TokenRegistry.t.sol
@@ -478,7 +478,7 @@ contract TokenRegistryTest is BridgeTestFixture {
         vm.assume(domain != homeDomain && domain != 0);
         address repr = tokenRegistry.getRepresentationAddress(domain, id);
         assertEq(repr, address(0));
-
+        expectTokenDeployedEmission(domain, id);
         address tokenAddress = tokenRegistry.exposed_deployToken(domain, id);
         BridgeToken token = BridgeToken(tokenAddress);
         repr = tokenRegistry.getRepresentationAddress(domain, id);
--- a/packages/contracts-bridge/contracts/test/TokenRegistry.t.sol
+++ b/packages/contracts-bridge/contracts/test/TokenRegistry.t.sol
@@ -96,12 +96,7 @@ contract TokenRegistryTest is BridgeTestFixture {
         bytes32 newId = bytes32("hey yoou");
         // It's the second contract that is been deployed by tokenRegistry
         // It deploys a bridgeToken during setUp() of BridgeTest
-        address calculated = computeCreateAddress(
-            address(tokenRegistry),
-            vm.getNonce(address(tokenRegistry))
-        );
-        vm.expectEmit(true, true, true, false);
-        emit TokenDeployed(newDomain, newId, calculated);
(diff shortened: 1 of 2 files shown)


packages/contracts-bridge/contracts/test/TokenRegistry.t.sol
```diff
@@ -18,7 +18,7 @@ contract TokenRegistryTest is BridgeTestFixture {
     using TypedMemView for bytes29;
     using BridgeMessage for bytes29;
 
-    function setUp() public override {
+    function setUp() public virtual override {
```
prestwich, line 21: if it just invokes super, this should actually be deleted

packages/contracts-bridge/contracts/test/TokenRegistry.t.sol
```diff
@@ -464,7 +471,10 @@ contract TokenRegistryTest is BridgeTestFixture {
         bytes32 id = "It's over 9000";
         address repr = tokenRegistry.getRepresentationAddress(domain, id);
         assertEq(repr, address(0));
-        address calculated = computeCreateAddress(address(tokenRegistry), 2);
+        address calculated = computeCreateAddress(
```
prestwich, line 474: should this be a named utility function? `expectedTokenAddress` or something like that?

packages/contracts-bridge/contracts/test/TokenRegistry.t.sol
odyslam: Makes sense

packages/contracts-bridge/contracts/test/TokenRegistry.t.sol
odyslam: TBH, I am not sold with adding all these utility functions in term of whether it makes the code more readable (albeit reducing code repeat).

prestwich: the reason I like utility functions is that it pushes the code towards even abstraction layers. When I'm reading a test, I don't want to break to remember how contract addresses are computed, or whether my registry uses `create` or `create2`. That is a different abstraction layer than the test functions on, and the details should be hidden when we're reading the test. Only things that work _on the same abstraction layer_ should be in the test

obv that's a bit of a muddy boundary, but clearly we're not interested in how addresses are calculated here

packages/contracts-bridge/contracts/test/TokenRegistry.t.sol
```diff
@@ -490,10 +478,7 @@ contract TokenRegistryTest is BridgeTestFixture {
         vm.assume(domain != homeDomain && domain != 0);
         address repr = tokenRegistry.getRepresentationAddress(domain, id);
         assertEq(repr, address(0));
-        address calculated = computeCreateAddress(address(tokenRegistry), 2);
-        // test event emmission
-        vm.expectEmit(true, true, true, false);
-        emit TokenDeployed(domain, id, calculated);
+
```
prestwich, line 481: was this removed on purpose?

odyslam: The tests should not be green in this one. Checking.

prestwich (approved):

PR #561, opened 2022-11-29
refactor: fork tests can easily be modified to any chain

anna-carroll:
## Motivation
We'd like the ability to run the fork tests on any chain

## Solution
Refactor to use CallBatch domain wherever possible so that the test domain can be changed in one single place in Reboot.t.sol.

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation

### DIFF
--- a/packages/contracts-ops/contracts/test/RebootEthereumBridgeRouter.t.sol
+++ b/packages/contracts-ops/contracts/test/RebootEthereumBridgeRouter.t.sol
@@ -19,31 +19,35 @@ contract EthereumBridgeRouterRebootTest is RebootTest, BridgeRouterTest {
     address bridgeRouterHarnessImpl;
     address tokenRegistryHarnessImpl;
 
-    string constant ethereum = "ethereum";
-
     function setUp() public override(BridgeRouterBaseTest, NomadTest) {
         setUpReboot("ethBridgeRouter");
+        require(
+            keccak256(bytes(localDomainName)) == keccak256(bytes("ethereum")),
+            "not ethereum bridge router"
+        );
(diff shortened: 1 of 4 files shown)


packages/contracts-ops/contracts/test/RebootEthereumBridgeRouter.t.sol
```diff
@@ -19,31 +19,35 @@ contract EthereumBridgeRouterRebootTest is RebootTest, BridgeRouterTest {
     address bridgeRouterHarnessImpl;
     address tokenRegistryHarnessImpl;
 
-    string constant ethereum = "ethereum";
-
     function setUp() public override(BridgeRouterBaseTest, NomadTest) {
         setUpReboot("ethBridgeRouter");
+        require(
+            keccak256(bytes(localDomainName)) == keccak256(bytes("ethereum")),
+            "not ethereum bridge router"
```
anna-carroll, line 26: i don't think there's a `skipTest` cheat code, so allowing it to fail at setup (thus not run any tests) with a clear error message

prestwich:
```
modifier skipNotEthereum() {
    if (keccak256(bytes(localDomainName)) == keccak256(bytes("ethereum"))) {
        _;
    }
}
```

anna-carroll: we don't have access to `localDomainName` in the inherited unit tests. also, that would mean that the unit tests don't run

PR #562, opened 2022-11-29
fix: governance tests work for governor and non-governor chains

anna-carroll:
## Motivation
If we run fork tests on other chains, the governance router tests need to function for both the governor domain and the non-governor domain

## Solution
for non-governor domains,
- for `handle` tests, ensure the remote domain is `setUp` to the governor domain. the GovernanceRouter tests assume that the remoteDomain is the governor domain; on the actual governor domain, the tests just transfer to the remote domain; but for non-governor domains, this is not possible
- for `TransferGovernor` tests, don't assume homeDomain is the current governor domain. 

## PR Checklist

- [ ] Added Tests
- [ ] Updated Documentation


anna-carroll:
Merge https://github.com/nomad-xyz/monorepo/pull/559 and https://github.com/nomad-xyz/monorepo/pull/560 first

odyslam (approved):

